PR #3, opened 2020-06-16
Add Azure Blob Store support for storage libs

kenoir:
For https://github.com/wellcomecollection/platform/issues/4397


storage/src/test/scala/uk/ac/wellcome/storage/store/StreamStoreTestCases.scala
```diff
@@ -63,24 +67,10 @@ trait StreamStoreTestCases[
...
-            val retrievedEntry = store.get(id).right.value
-
-            assertEqualT(initialEntry, retrievedEntry.identifiedT)
-          }
-        }
-      }
     }
 
     describe("put") {
-      it("can put a stream without metadata") {
```
kenoir, line 83: Removing test dupes leftover from removal of metadata storage from `StreamStore`

storage/src/main/scala/uk/ac/wellcome/storage/store/azure/AzureStreamReadable.scala
```diff
@@ -0,0 +1,60 @@
...
+import uk.ac.wellcome.storage._
+import uk.ac.wellcome.storage.store.Readable
+import uk.ac.wellcome.storage.streaming.InputStreamWithLength
+import scala.util.{Failure, Success, Try}
+
+trait AzureStreamReadable
+    extends Readable[ObjectLocation, InputStreamWithLength]
+    with Logging {
+  implicit val blobClient: BlobServiceClient
+  val maxRetries: Int
```
alexwlchan, line 16: Maybe push the retry logic into an AzureGet object, similar to what I've done with S3Get? We'll almost certainly want to reuse most of this.

storage/src/main/scala/uk/ac/wellcome/storage/store/azure/AzureStreamWritable.scala
```diff
@@ -0,0 +1,31 @@
...
+    Try {
+      val individualBlobClient =
+        blobClient
+          .getBlobContainerClient(location.namespace)
+          .getBlobClient(location.path)
+
+      individualBlobClient.upload(
+        inputStream,
+        inputStream.length,
+        true
```
alexwlchan, line 25: The unexplained boolean is an issue, and the Azure SDK doesn't have a neater comment for it. A comment explaining:

1. What this parameter does
2. Why we're setting this value

would be useful.

storage/src/test/scala/uk/ac/wellcome/storage/store/StreamStoreTestCases.scala
```diff
@@ -91,45 +81,38 @@ trait StreamStoreTestCases[
...
-            length = bytes.length + 1
-          )
-
-          withStoreImpl(initialEntries = Map.empty) { store =>
-            val result = store.put(createId)(brokenStream).left.value
-
-            result shouldBe a[IncorrectStreamLengthError]
+      if (skipStreamLengthTests) {
+        it("skips stream length tests") {}
+      } else {
```
alexwlchan, line 86: Scalatest has some nicer mechanisms for programatically skipping tests: https://www.scalatest.org/user_guide/tagging_your_tests

kenoir: I've used `assume` rather than tags as I don't think there is a nice way to conditionally tag individual tests (with tags it seems you tell the runner which to execute which would not easily fit this use case).

storage/src/main/scala/uk/ac/wellcome/storage/store/azure/AzureStreamReadable.scala
```diff
@@ -0,0 +1,39 @@
...
+    buildInputStream(blobInputStream)
+  }
+
+  private def buildInputStream(
+    blobInputStream: BlobInputStream): InputStreamWithLength = {
+    new InputStreamWithLength(
+      blobInputStream,
+      length = blobInputStream.getProperties.getBlobSize,
+    )
+  }
```
alexwlchan, line 29: I’d inline this function; breaking it out isn’t adding anything.

storage/src/main/scala/uk/ac/wellcome/storage/store/RetryableReadable.scala
```diff
@@ -0,0 +1,39 @@
...
+  def retryableGet(location: ObjectLocation): Either[ReadError, T] =
+    getOnce().retry(maxRetries)(location)
+
+  private def getOnce(): ObjectLocation => Either[ReadError, T] =
+    (location: ObjectLocation) =>
+      Try {
+        retryableGetFunction(location)
+      } match {
+        case Success(t)   => Right(t)
+        case Failure(err) => Left(buildGetError(err))
```
alexwlchan, line 37: Thinking back to the issues we’ve been having with the tagger, `buildGetError` should get the location as well as the error. That way we can log the exception and the location together.

storage/src/main/scala/uk/ac/wellcome/storage/store/azure/AzureStreamWritable.scala
```diff
@@ -0,0 +1,43 @@
...
+      val individualBlobClient =
+        blobClient
+          .getBlobContainerClient(location.namespace)
+          .getBlobClient(location.path)
+
+      individualBlobClient.upload(
+        inputStream,
+        inputStream.length.toLong,
+        allowOverwrites
+      )
```
alexwlchan, line 33: 🎉

storage/src/main/scala/uk/ac/wellcome/storage/store/azure/AzureStreamWritable.scala
```diff
@@ -0,0 +1,43 @@
...
+
+    } match {
+      case Success(_) => Right(Identified(location, inputStream))
+      case Failure(
+          error: BlobStorageException
+          ) if error.getErrorCode == BlobErrorCode.BLOB_ALREADY_EXISTS =>
```
alexwlchan, line 39: Having named error codes is really nice. 👍

alexwlchan: nit: `exc: BlobStorageException` instead of `err`?

storage/src/test/scala/uk/ac/wellcome/storage/store/azure/AzureStreamStoreTest.scala
```diff
@@ -0,0 +1,82 @@
...
+    initialEntries.foreach {
+      case (location, data) =>
+        azureClient
+          .getBlobContainerClient(location.namespace)
+          .getBlobClient(location.path)
+          .upload(data, data.length)
+
+        azureClient
+          .getBlobContainerClient(location.namespace)
+          .getBlobClient(location.path)
```
alexwlchan, line 48: What’s this `getBlobClient` call doing for us?

storage/src/test/scala/uk/ac/wellcome/storage/store/azure/AzureStreamStoreTest.scala
```diff
@@ -0,0 +1,82 @@
...
+
+        withStoreImpl(
+          initialEntries = Map(id -> entry),
+          storeContext = AzureStreamStoreContext(
+            allowOverwrite = Some(false)
+          )
+        ) { store =>
+          val result = store.put(id)(entry).left.value
+
+          result shouldBe a[OverwriteError]
```
alexwlchan, line 77: This test confirms that you have an error, but it doesn’t tell us whether the flag was respected.

The test should try to write different contents to the location, then read it back and check it contains the original contents.

storage/src/main/scala/uk/ac/wellcome/storage/store/RetryableReadable.scala
kenoir: I worked around this by logging here!

PR #5, opened 2020-06-16
Make visibility timeout optional

warrd:
Localstack doesn't like specifying a visibility timeout so make it optional.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,9 @@
+RELEASE_TYPE: patch
+
+### Libraries affected
+
+`messaging`
+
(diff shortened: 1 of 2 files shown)


alexwlchan:
I’m not a huge fan of muddying up this interface for the benefit of localstack.

Given that localstack isn’t used anywhere in this repo (or we’d have caught it), how would you feel about instead deleting all the localstack code from this repo, and inlining the bits we actually use in the catalogue repo?

warrd:
Works for me, as the locallstack stuff is only used in one test class anyway

PR #8, opened 2020-06-17
Tags should adhere to Readable

kenoir:
`RetryableGet` has been replaced in favour of `RetryableReadable`. 

A new trait `Updatable` has been made available and is used in `Tags`.

```scala
trait Updatable[Ident, T] {
  type UpdateEither = Either[UpdateError, Identified[Ident, T]]
  type UpdateFunction = T => Either[UpdateFunctionError, T]
  def update(id: Ident)(updateFunction: UpdateFunction): UpdateEither
}
```

`Tags` now meets the interfaces of  to `RetryableReadable` and `Updatable`.

Consumers will need to update code where `update` and `get` have been used on `S3Tags`.

---

Will require downstream updates tp the `bag_tagger` & `bag_verifier` services in https://github.com/wellcomecollection/storage-service.

Part of: https://github.com/wellcomecollection/platform/issues/4397


storage/src/main/scala/uk/ac/wellcome/storage/store/RetryableReadable.scala
```diff
@@ -33,7 +27,10 @@ trait RetryableGet[T] {
       Try {
         retryableGetFunction(location)
       } match {
-        case Success(t)   => Right(t)
-        case Failure(err) => Left(buildGetError(err))
+        case Success(t) => Right(t)
+        case Failure(err) =>
+          val error = buildGetError(err)
+          warn(s"$error when trying to get $location")
```
alexwlchan, line 33: Let's swap this around slightly:

```suggestion
          warn(s"Error when trying to get $location: $error")
```

The error could be much longer than the location. Having them all neatly line up will make it easier to spot commonalities if you have a stack of logs next to each other.

storage/src/main/scala/uk/ac/wellcome/storage/tags/Tags.scala
```diff
@@ -25,17 +32,17 @@ trait Tags[Ident] extends Logging {
...
+      result <- if (newTags == existingTags.identifiedT) {
         debug(s"Tags on $id: no change, so skipping a write")
         Right(existingTags)
       } else {
         debug(s"Tags on $id: tags have changed, so writing new tags")
-        put(id = id, tags = newTags) match {
-          case Right(value) => Right(value)
+        writeTags(id = id, tags = newTags) match {
+          case Right(value) => Right(Identified(id, value))
           case Left(err)    => Left(UpdateWriteError(err))
```
alexwlchan, line 46: Worth a similar warning log here?

kenoir: Will apply in next PR.

PR #12, opened 2020-06-18
Fixes & Tests for EnrichConfig

kenoir:
Updates `EnrichConfig` to add:

- `getStringOption`
- `requireString`
- `getIntOption`
- `requireInt`

Deprecates:

- `get[T]`
- `getOrElse[T]`
- `required[T]`

This change is to provide working extraction of optional `Int` & `String` config.


typesafe_app/src/main/scala/uk/ac/wellcome/typesafe/config/builders/EnrichConfig.scala
```diff
@@ -3,31 +3,111 @@ package uk.ac.wellcome.typesafe.config.builders
...
+      System.exit(1)
+    }
+
+    def getStringOption(path: String): Option[String] = {
+      pathExists(path) { cleanPath =>
+        underlying.getString(cleanPath)
+      }
+    }
+
+    def getStringOrElse(path: String)(default: String): String = {
```
jamieparkinson, line 34: Is it necessary for us to provide the `getOrElse` methods given we provide ones that return an Option? If we are to provide them, why don't we use the `getOption` method inside them rather than duplication the getting code?

typesafe_app/src/main/scala/uk/ac/wellcome/typesafe/config/builders/EnrichConfig.scala
```diff
@@ -3,31 +3,111 @@ package uk.ac.wellcome.typesafe.config.builders
...
+    }
+
+    def getIntOrElse(path: String)(default: Int): Int = {
+      pathExists(path) { cleanPath =>
+        underlying.getInt(cleanPath)
+      } getOrElse(default)
+    }
+
+    def requiredInt(path: String): Int = {
+      val result = pathExists(path) { cleanPath =>
```
jamieparkinson, line 68: Same sort of comment re duplication here - why not `val result = getIntOption(path)`?

typesafe_app/src/main/scala/uk/ac/wellcome/typesafe/config/builders/EnrichConfig.scala
kenoir: Probably not, but I want parity with the existing functions.

kenoir: On reflection - I will remove them - I agree with your suggestion, and i'm deprecating rather than removing the old functions.

typesafe_app/src/main/scala/uk/ac/wellcome/typesafe/config/builders/EnrichConfig.scala
```diff
@@ -3,31 +3,111 @@ package uk.ac.wellcome.typesafe.config.builders
...
+      }
+    }
+
+    def getStringOrElse(path: String)(default: String): String = {
+      pathExists(path) { cleanPath =>
+        underlying.getString(cleanPath)
+      } getOrElse(default)
+    }
+
+    def requiredString(path: String): String = {
```
warrd, line 40: What about `requireString` or `getString`? `requiredString` is read more like a noun which is a bit confusing in this context.

Also I would be more up for returning a `Either` / `Try` than throwing.

kenoir: This calls system.exit, so the throw here is just to make the compiler happy. Agree about the naming.

typesafe_app/src/main/scala/uk/ac/wellcome/typesafe/config/builders/EnrichConfig.scala
```diff
@@ -3,31 +3,87 @@ package uk.ac.wellcome.typesafe.config.builders
...
 
 object EnrichConfig {
+
   implicit class RichConfig(val underlying: Config) extends AnyVal {
+    private def cleanUpPath(p: String): String =
+    // Sometimes we may get a path that features two double dots, if there's an
+    // empty namespace -- in this case, elide the two dots into one.
+      p.replaceAllLiterally("..", ".")
+
+    private def pathExists[T](path: String)(f: String => T): Option[T] = {
```
alexwlchan, line 13: minor: the name of this sounds more like something that returns a boolean – does the path exist? I'd prefer something like `getPathValue`.

typesafe_app/src/main/scala/uk/ac/wellcome/typesafe/config/builders/EnrichConfig.scala
kenoir: I'm not sure that calling `system.exit` is the correct thing to do, but it's a large behaviour change to remove that.

kenoir:
Resolves: https://github.com/wellcomecollection/platform/issues/3824

PR #14, opened 2020-06-22
Split ObjectLocation into per-provider types

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4596


storage/src/main/scala/uk/ac/wellcome/storage/Location.scala
```diff
@@ -0,0 +1,9 @@
+package uk.ac.wellcome.storage
+
+trait Location {
+  def join(parts: String*): Location
```
kenoir, line 4: Why does this need `join`?

alexwlchan: Good question.

storage/src/main/scala/uk/ac/wellcome/storage/s3/S3ObjectLocation.scala
```diff
@@ -0,0 +1,65 @@
...
+    )
+}
+
+// Note: the precursor to these classes was ObjectLocation, which used the
+// fields "namespace/path" rather than provider-specific terms.
+//
+// These decoders allow you to read JSON that was encoded with the old code.
+
+trait S3Decodable {
+  def createDecoder[T](constructor: (String, String) => T): Decoder[T] =
```
kenoir, line 31: 💃

PR #16, opened 2020-06-23
Break out project builds

kenoir:


.travis.yml
```diff
@@ -30,22 +29,60 @@ env:
...
-      services:
-        - docker
+      name: fixtures
+      script: sbt "project fixtures" test
+    - name: json
+      script: sbt "project json" test
+    - name: messaging
+      script: sbt "project messaging" dockerComposeUp test dockerComposeStop
+    - name: messaging_typesafe
+      script: sbt "project messaging_typesafe" test
```
alexwlchan, line 50: Historically these ran as part of a single task (in their per-repo build). Worth consolidating them to cut down on build times a bit? You'll still get a strong hint of where the issue is.

.travis.yml
```diff
@@ -30,22 +29,60 @@ env:
...
+    # Publish
+    - stage: publish
+      name: fixtures
+      script: sbt "project fixtures" publish
+    - name: json
+      script: sbt "project json" publish
+    - name: messaging
+      script: sbt "project messaging" publish
+    - name: messaging_typesafe
+      script: sbt "project messaging_typesafe" publish
```
alexwlchan, line 72: Ditto above. Multiplying the number of Travis VMs we need to provision has been historically not-awesome for build times.

RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Split build steps for performance.
```
alexwlchan, line 3: Let’s be clear that this is a build system change, not something that affects the libraries:

```suggestion
No user-facing changes. Split the build steps in CI for the scala-libs repo to improve […]
```

Also, is this actually improving performance (and if so, of what)? I thought it was more about making test failures easier to debug?

.travis.yml
kenoir: This build seems to take about the same time as it did before.

RELEASE.md
kenoir: It would improve performance if we had higher parallelism I think, but I take the point.

.travis.yml
kenoir: I do agree it makes sense to consolidate these though.

PR #18, opened 2020-07-13
Store Instants in Dynamo as timestamps in seconds

jamieparkinson:
Does what it says on the tin so that Dynamo can remove expired rows

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,9 @@
+RELEASE_TYPE: minor
+
+### Libraries affected
+
+`storage`
+
+### Description
+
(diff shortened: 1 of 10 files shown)


kenoir:
Are you sure this will not impact deserialising from existing data stores that have an `Instant` stored in the previous format?

jamieparkinson:
It will break that, but it's only used in the locking service so *in theory* that's not a huge problem. Do you know where else this is used / should I make this a major semver release?

kenoir:
I believe this will break deserialising ingests in the storage service if we were to use this code. I would prefer this were implemented so that clients can opt into the "second" rather than "millisecond" deserialisation.

PR #19, opened 2020-07-23
Add the base Location classes plus compatibility shims

alexwlchan:
Starting to bring across bits of https://github.com/wellcomecollection/platform/issues/4596

This includes compatibility shims so you can read from an existing data store that uses ObjectLocation[Prefix] with S3ObjectLocation[Prefix], and it should transparently decode the old-style data for you.

### DIFF
--- a/storage/src/test/scala/uk/ac/wellcome/storage/s3/S3ObjectLocationTest.scala
+++ b/storage/src/test/scala/uk/ac/wellcome/storage/s3/S3ObjectLocationTest.scala
@@ -1,0 +1,137 @@
+package uk.ac.wellcome.storage.s3
+
+import org.scalatest.funspec.AnyFunSpec
+import org.scalatest.matchers.should.Matchers
+import org.scanamo.auto._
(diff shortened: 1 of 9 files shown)


storage/src/test/scala/uk/ac/wellcome/storage/s3/S3ObjectLocationTest.scala
```diff
@@ -0,0 +1,137 @@
...
+
+      it("can store a location in DynamoDB") {
+        withLocalDynamoDbTable { table =>
+          putTableItem(table = table, item = item)
+
+          getTableItem[IdentifiedLocation](id = item.id, table = table).get.right.value shouldBe item
+        }
+      }
+
+      it("can retrieve an old-style location from DynamoDB") {
```
kenoir, line 60: 👏

alexwlchan: I intend to be fairly aggressive about expunging them, but I figure we can't flip the entire catalogue at once!

PR #33, opened 2020-08-18
Encapsulate the DynamoFormat[ExpiringLock] in the lock DAO

jamieparkinson:
I've put this as a major release because it is technically a breaking change (the implicit will become unused in client code) but could be persuaded to downgrade it


jamieparkinson:
The unused implicit in scope breaks the build but yeah, it's simple enough that I think minor is fine

PR #35, opened 2020-08-19
Retry the writeTags operation in S3Tags

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4739

### DIFF
--- a/storage/src/main/scala/uk/ac/wellcome/storage/tags/s3/S3Tags.scala
+++ b/storage/src/main/scala/uk/ac/wellcome/storage/tags/s3/S3Tags.scala
@@ -1,21 +1,11 @@
 package uk.ac.wellcome.storage.tags.s3
 
 import com.amazonaws.services.s3.AmazonS3
-import com.amazonaws.services.s3.model.{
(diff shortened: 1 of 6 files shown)


storage/src/main/scala/uk/ac/wellcome/storage/tags/s3/S3Tags.scala
```diff
@@ -50,7 +57,14 @@ class S3Tags(val maxRetries: Int = 3)(implicit s3Client: AmazonS3)
         )
       )
     } match {
-      case Success(_)   => Right(tags)
+      case Success(_) => Right(tags)
+
+      case Failure(exc: AmazonS3Exception)
+          if exc.getMessage.startsWith(
+            "We encountered an internal error. Please try again.") ||
+            exc.getMessage.startsWith("Please reduce your request rate.") =>
```
alicefuzier, line 65: Is it not possible to match on status code 500 or 5xx instead of the error message? Seems a little flaky matching on the error message, which could change in theory

alicefuzier (approved):

PR #36, opened 2020-08-20
Drop warning logs that listing a prefix w/o trailing slash may be error

alexwlchan:


alexwlchan:
Discussed in Slack with @alicefuzier.

PR #43, opened 2020-09-15
Retry timeouts from Azure

alicefuzier:


storage/src/main/scala/uk/ac/wellcome/storage/azure/AzureStorageErrors.scala
```diff
@@ -1,12 +1,19 @@
...
+  RetryableError,
+  StoreReadError
+}
 
 object AzureStorageErrors {
   val readErrors: PartialFunction[Throwable, ReadError] = {
     case exc: BlobStorageException if exc.getStatusCode == 404 =>
       DoesNotExistError(exc)
+    case exc if exc.getMessage.contains("TimeoutException") =>
+      new StoreReadError(exc) with RetryableError
```
alexwlchan, line 16: Maybe worth a comment/example of the error we're trying to catch? Here's one from the logs:

> reactor.core.Exceptions$ReactiveException: java.util.concurrent.TimeoutException: Did not observe any item or terminal signal within 60000ms in 'map' (and no fallback has been configured)

Normally we'd match on the exception class rather than the message, but this exception is somewhat weird and I don't think we can do that.

storage/src/test/scala/uk/ac/wellcome/storage/azure/AzureStorageErrorsTest.scala
```diff
@@ -0,0 +1,14 @@
...
+import org.scalatest.matchers.should.Matchers
+import uk.ac.wellcome.storage.RetryableError
+
+class AzureStorageErrorsTest extends AnyFunSpec with Matchers {
+  it("marks a timeout as a retriable error"){
+    val exception = new Exception("something TimeoutException something")
+    val error = AzureStorageErrors.readErrors(exception)
+    error shouldBe a[RetryableError]
+    error.e shouldBe exception
+  }
```
alexwlchan, line 13: 👍

storage/src/main/scala/uk/ac/wellcome/storage/azure/AzureStorageErrors.scala
alicefuzier: yeah, we can't unfortunately. Even reproducing that in tests is somewhat fiddly cause it's a path-dependent type or whatever they're called so I think this is the most robust way of catching that actually. 

Forgot the comment, you're right. I had it in the other PR and then deleted it instead of copying it across 🤦‍♀️

alexwlchan (approved):

PR #45, opened 2020-09-22
Make validity of pre-signed URL a config flag

alicefuzier:


alexwlchan (approved):
Nice changelog entry! ✨

alexwlchan:
Is it worth a test that this flag is being used? Create a pre-signed URL with a 0-second expiry? (Does that even make sense/does the S3 container support that?)

alicefuzier:
mmh I can play around a bit but I wouldn't be surprised if the s3 container just ignored it

alexwlchan:
I wonder what we do in the S3Uploader tests? 🤔

alexwlchan:
So the container does pay attention to it, and we do test it: https://github.com/wellcomecollection/scala-libs/blob/dc6f6a0b35f46d95d1fadfc199c9b7117a667563/storage/src/test/scala/uk/ac/wellcome/storage/services/s3/S3UploaderTest.scala#L118-L148

You might struggle to make a reliable test for this if it demands non-zero expiry time – because it's gonna go straight from "get the URL" to "put the block" – unless you make the blockSize a flag, and turn it way down so we have to make dozens of Azure requests.

alicefuzier:
mmh so, it's a bit hard to test in its entirety because Azurite doesn't support the PutBlockFromUrl thing. Since the flag behaviour is already tested in the S3Uploader, I'm just going to change the AzurePutBlockFromUrlTransfer to have the S3Uploader injected and then check the interactions

alexwlchan (approved):

PR #46, opened 2020-09-30
Retry errors from s3 in S3Transfer

alicefuzier:


storage/src/main/scala/uk/ac/wellcome/storage/transfer/s3/S3Transfer.scala
```diff
@@ -96,10 +88,14 @@ class S3Transfer(implicit s3Client: AmazonS3)
...
-  private def getStream(location: S3ObjectLocation): Try[S3ObjectInputStream] =
-    Try {
+  private def getStream(location: S3ObjectLocation) = {
+    val errorOrStream = Try {
       s3Client.getObject(location.bucket, location.key)
-    }.map { _.getObjectContent }
+    }.toEither.left.map(error => S3Errors.readErrors(error)).map {
+      _.getObjectContent
+    }
+    errorOrStream.retry(maxAttempts = 3)
```
alexwlchan, line 97: Maybe we could replace this with an S3StreamReadable or S3StreamStore? That would give us the retrying functionality “for free” and is already the correct type.

The reason this code wasn't using it before is because S3StreamStore returns an Either[_] rather than a Try[_], and I didn't think it worth doing an Either -> Try conversion. Doing a Try -> Either conversion is a bit icky as well.

(If you inject an S3StreamStore, I think you can also ditch the implicit S3 client from the constructor, which IMO is clearer. Right now one argument can be derived from the other, which seems weird.)

storage/src/main/scala/uk/ac/wellcome/storage/transfer/s3/S3Transfer.scala
```diff
@@ -112,20 +108,47 @@ class S3Transfer(implicit s3Client: AmazonS3)
...
-        case Failure(err) => Left(TransferDestinationFailure(src, dst, err))
-      }
+      transfer <- tryCopyFromSource(src, dst, copyRequest)
+        .retry(maxAttempts = 3)
+        .left
+        .map(err => TransferSourceFailure(src, dst, err.e))
+      result <- tryCopyToDestination(src, dst, transfer)
+        .retry(maxAttempts = 3)
+        .left
+        .map(err => TransferDestinationFailure(src, dst, err.e))
```
alexwlchan, line 118: ✨ Retrying aside, this is a much cleaner way of handling this for comprehension.

storage/src/main/scala/uk/ac/wellcome/storage/transfer/s3/S3Transfer.scala
```diff
@@ -112,20 +108,47 @@ class S3Transfer(implicit s3Client: AmazonS3)
...
     } yield result
   }
+
+  private def tryCopyToDestination(src: S3ObjectLocation,
+                                   dst: S3ObjectLocation,
+                                   transfer: Copy) = {
```
alexwlchan, line 124: Minor: maybe we could rename this type `TransferManagerCopyRequest` when we import it? It took me a moment to figure out what this type was, because "Copy" is so generic.

storage/src/main/scala/uk/ac/wellcome/storage/transfer/s3/S3Transfer.scala
```diff
@@ -112,20 +108,47 @@ class S3Transfer(implicit s3Client: AmazonS3)
...
+  }
+}
+
+object S3Transfer {
+  def apply(implicit s3Client: AmazonS3) = {
+    val transferManager: TransferManager = TransferManagerBuilder.standard
+      .withS3Client(s3Client)
+      .build
+    new S3Transfer(transferManager)
+  }
```
alexwlchan, line 153: ✨

storage/src/test/scala/uk/ac/wellcome/storage/transfer/s3/S3PrefixTransferTest.scala
```diff
@@ -88,7 +89,9 @@ class S3PrefixTransferTest
   ): R = {
     implicit val listing: S3ObjectLocationListing = S3ObjectLocationListing()
 
-    implicit val transfer: S3Transfer = new S3Transfer() {
+    implicit val transfer: S3Transfer = new S3Transfer(TransferManagerBuilder.standard
+      .withS3Client(s3Client)
+      .build) {
```
alexwlchan, line 94: Minor: this is quite a lot on one line. Maybe define the TransferManager on its own line?

storage/src/test/scala/uk/ac/wellcome/storage/transfer/s3/S3TransferTest.scala
```diff
@@ -92,4 +106,36 @@ class S3TransferTest
...
+        withContext { implicit context =>
+          withSrcStore(initialEntries = Map(src -> t)) { srcStore =>
+            withDstStore(initialEntries = Map.empty) { dstStore =>
+
+              val failingOnceTransfer = mock[TransferManager]
+              when(failingOnceTransfer.copy(any[CopyObjectRequest]))
+                .thenThrow(s3ServerException)
+                .thenAnswer((invocation: InvocationOnMock) => {
+                transferManager.copy(invocation.getArguments.toList.head.asInstanceOf[CopyObjectRequest])
+              })
```
alexwlchan, line 126: Ooh, I’d wondered how you fail once and then call the underlying method. Nice!

storage/src/test/scala/uk/ac/wellcome/storage/transfer/s3/S3TransferTest.scala
```diff
@@ -92,4 +106,36 @@ class S3TransferTest
...
+
+              result.right.value shouldBe TransferPerformed(src, dst)
+
+              srcStore.get(src).right.value.identifiedT shouldBe t
+              dstStore.get(dst).right.value.identifiedT shouldBe t
+            }
+          }
+        }
+      }
+    }
```
alexwlchan, line 140: Normally for retryable logic I’d consider a test that we do eventually give up – if I have a TransferManager that repeatedly returns a 500 error, it returns a Left[_] rather than spinning forever.

alexwlchan (approved):
Really nice work. You’ve made the internals of S3Transfer nicer to follow, and that’s before you get to the retrying logic.

A couple of suggestions inline, but nothing substantial.

storage/src/main/scala/uk/ac/wellcome/storage/transfer/s3/S3Transfer.scala
alicefuzier: Why is a Try -> Either (or the other way around) a bit icky?

alicefuzier: Your comment makes sense though. Will do

storage/src/main/scala/uk/ac/wellcome/storage/transfer/s3/S3Transfer.scala
alicefuzier: It's not a copy request though. CopyRequest is what gets passed to `transfermanager.copy` I think. Maybe `TransferManagerCopyResponse`?

alexwlchan: Aha, then I was double-confused. `TransferManagerCopyResponse` is good.

storage/src/main/scala/uk/ac/wellcome/storage/transfer/s3/S3Transfer.scala
alexwlchan: > Why is a Try -> Either (or the other way around) a bit icky?

I always find it a bit odd if you create an Either/Try then immediately cast to something else in the same line, because why not go straight to the thing you wanted? But it might just be a style thing.

alicefuzier: In this case you can't go straight to `Either` because `s3Client.getObject(location.bucket, location.key)` thows an Exception and `Try` is the only thing that allows you to deal with it. I find that when we deal with Java libraries, where the pattern is usually to thow and let the exception bubble up, then we have to go via `Try`. `Either` and `Try` have a lot in common but aren't quite the same and sometimes only one of the 2 will do, so I don't really see anything wrong with going back and forth and using the one that is more convenient in that moment

storage/src/main/scala/uk/ac/wellcome/storage/transfer/s3/S3Transfer.scala
alicefuzier: Actually even`TransferManagerCopyResponse` isn't really good because `Copy.waitForCopyResult` returns a `CopyResult` so that would be also very confusing. I'd rather not mess with the names here. At least if you leave the aws type you can follow that through and read the comment in the aws sdk for `Copy`. if we rename it, we add one extra level of indirection without really explaining what it is

PR #52, opened 2020-11-19
Add support for eventually consistent reads

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4873

PR #53, opened 2020-11-19
Fix a flaky test in S3TagsTest

alexwlchan:
We should create 1 less than the max tags, not max tags -- otherwise we'll be unable to append tags in the S3 test.

This is breaking #52.


alexwlchan:
You can see an example of this failure at https://buildkite.com/wellcomecollection/scala-libraries/builds/89#2e703177-5890-40c0-bc1e-6204914d2517

kenoir (approved):

PR #54, opened 2020-11-20
Allow consistency to be set from DynamoSingleVersionStore, DynamoMultipleVersionStore

kenoir:
Allows consistency mode to be set in config, in turn allowing DynamoSingleVersionStore, DynamoMultipleVersionStore to have more easily configured consistency.

Helps with https://github.com/wellcomecollection/platform/issues/4873

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: minor
+
+Allow consistency to be set from DynamoSingleVersionStore, DynamoMultipleVersionStore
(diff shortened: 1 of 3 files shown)


kenoir:
I looked for a way to better test this, but I ended up concluding that this compiles and existing tests pass is sufficient.

alexwlchan (commented):
I'm not sure this is right – it's a property of the Store, not the Table.

kenoir:
> I'm not sure this is right – it's a property of the Store, not the Table.

Don't understand? The config is being set on the store?

alexwlchan:
>> I'm not sure this is right – it's a property of the Store, not the Table.
>
> Don't understand? The config is being set on the store?

It looks to me like you're adding a third field to DynamoConfig?

kenoir:
> It looks to me like you're adding a third field to DynamoConfig?

Ah I see - my expectation was that `DynamoConfig` was for the store and not the table - but I think you are saying we should scope it to the table only?

alexwlchan (approved):
Yup, this is what I was suggesting.

PR #57, opened 2020-12-02
Use GetItem instead of Query in the DynamoDB stores

alexwlchan:
It's hard to be sure, but GetItem cannot be slower than Query (and I suspect involves a bit less work on DynamoDB's side). Given we can swap it out and everything works the same, we might as well.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Improve the efficiency of reading items from the Dynamo stores.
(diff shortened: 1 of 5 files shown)


kenoir (approved):

alicefuzier (approved):

kenoir:
Relevant: https://forums.aws.amazon.com/thread.jspa?threadID=93743, but we should do this anyway as it makes more sense when reading the code this way.

PR #58, opened 2020-12-07
Make a single call to DynamoDB in getLatest(), not two

alexwlchan:
Previously, getLatest() was implemented in an inefficient way:

1. What’s the max version for a given ID?
2. What's the item for this given (id, version) pair?

We have to look up the whole DynamoDB item to answer question (1), so just return that rather than dropping the rest of the item to return an Int.

This doesn't change the interface to getLatest(); it just becomes more efficient. It does change the interface of Maxima (hence the major version bump), but we only use that in one place outside the libraries, and it's accompanied by a comment describing this exact change: https://github.com/wellcomecollection/storage-service/blob/72aad5d5edc6dc125d079d055643ea11dc4d94c7/bag_versioner/src/main/scala/uk/ac/wellcome/platform/storage/bag_versioner/versioning/dynamo/DynamoIngestVersionManagerDao.scala#L79-L85

Another piece for https://github.com/wellcomecollection/platform/issues/4913. I don't have enough profiling data to say how much of a difference this will make (if any), but it's hard to imagine shedding several million DynamoDB queries won't have an impact on speed (and cost!).

This function is used plenty in the catalogue:

* It's used to look up source records in the Calm, METS and Sierra transformers
* It's used to look up records in the Items VHS in the Sierra item merger, and then again (via VersionedStore.update() to actually write the record)
* It's used to upsert records in the bib merger, item merger, and items_to_dynamo
* It's used to decide if we should replace a record in the Calm adapter

Similarly in the storage service:

* The bag and ingest trackers both use it to retrieve and update records
* The replica aggregator uses it to upsert replica records

So halving the number of Dynamo calls it makes will hopefully make a tidy dent in the pipeline performance.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,6 @@
+RELEASE_TYPE: major
+
+This release changes the behaviour of Maxima to return an `Identified[_, _]`, rather than a single value.
+This reduces doing double-lookups if you want to get the max version of something in a store, and retrieve the whole record rather than just the max version.
+
+The practical upshot for our code is that if you call `getLatest(…)`, now you're making a single request to DynamoDB rather than two.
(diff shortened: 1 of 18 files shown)


warrd (approved):

warrd:
Do we have buildkite enabled for this? Can't see the integration :thinking:

alexwlchan:
Yup: https://buildkite.com/wellcomecollection/scala-libraries/builds/112

PR #72, opened 2021-01-24
Don't acquire known-unusable locks in LockingService

alexwlchan:
The DynamoDB lock table is now one of the main costs of a reindex in the catalogue pipeline, so I've been looking for ways to make it more efficient.  We do see locking failures in the matcher logs, so some processes are trying to acquire the same/similar locks.

Previously, the locking service would lock IDs in sequence, e.g.:

    lockDao.lock(id1)
    lockDao.lock(id2)
    lockDao.lock(id3)

But if we get a failure on locking `id1`, trying to lock `id2` and `id3` is pointless.  We create and then delete two locks that we know we can never use, which is expensive and potentially stops another process from locking one of those IDs.

Now, if the locking service detects a failure to lock `id1`, it skips locking `id2` and `id3`, and goes straight to the cleanup stage.

I don't know how much of a difference this will make -- I'd bet a few percentage points at best -- because this code is already fairly heavily optimised, but having spotted this simple optimisation I figured it couldn't hurt to add.

### DIFF
--- a/storage/src/main/scala/uk/ac/wellcome/storage/locking/LockingService.scala
+++ b/storage/src/main/scala/uk/ac/wellcome/storage/locking/LockingService.scala
@@ -62,7 +62,36 @@ trait LockingService[Out, OutMonad[_], LockDaoImpl <: LockDao[_, _]]
     */
   private def getLocks(ids: Set[lockDao.Ident],
                        contextId: lockDao.ContextId): LockingServiceResult = {
-    val lockResults = ids.map { lockDao.lock(_, contextId) }
+
+    // We lock the IDs one-by-one, but if any ID fails to lock, we skip
+    // even trying to lock the remaining IDs.
+    //
+    // This reduces the amount of churn in the LockDao: we're skipping creating
+    // and deleting a lock we know will never be used.
+    val lockResults =
+      ids.foldLeft(Set[lockDao.LockResult]())(
+        (existingResults: Set[lockDao.LockResult], id) => {
+          val failedIds = getFailedLocks(existingResults).map { _.id }
+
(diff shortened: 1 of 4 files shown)


storage/src/main/scala/uk/ac/wellcome/storage/locking/LockingService.scala
```diff
@@ -62,7 +62,38 @@ trait LockingService[Out, OutMonad[_], LockDaoImpl <: LockDao[_, _]]
     */
   private def getLocks(ids: Set[lockDao.Ident],
                        contextId: lockDao.ContextId): LockingServiceResult = {
-    val lockResults = ids.map { lockDao.lock(_, contextId) }
+
+    // We lock the IDs one-by-one, but if any ID fails to lock, we skip
```
jamieparkinson, line 66: I think this can be done lazily using recursion:

```scala
  @tailrec
  private def getLocks(ids: Set[lockDao.Ident],
                       contextId: lockDao.ContextId): LockingServiceResult =
    if (ids.isEmpty) {
      Right(contextId)
    } else {
      lockDao.lock(ids.head, contextId) match {
        case Left(failure) =>
          Left(FailedLock(contextId, ids.tail.map(skipped) + failure))
        case Right(_) => getLocks(ids.tail, contextId)
      }
    }

  private def skipped[Ident](id: Ident) =
    LockFailure(
      id,
      e = new Throwable(
        s"Skipping locking $id; other IDs have already failed to lock"
      )
    )
```

Where you can replace pattern matching with mapping or vice versa as per your taste, and where `getFailedLocks` has been removed

kenoir (approved):

PR #73, opened 2021-02-02
Add some tests to describe dynamo formatting behaviour

jamieparkinson:
Wrote these to see what happens, they're useful documentation


alexwlchan (approved):
Better documentation than Scanamo itself. 👍

PR #74, opened 2021-02-02
Add ability to run an arbitrary graph between a provided sink and source in SQSStream

jamieparkinson:
I've wanted this for ages as having to propagate messages all the way through `runStream` can be a pain - thought I'd finally implement it.

Currently, `SQSStream` offers us
- `foreach` to apply a basic transformation function returning a `Future`
- `runStream` to modify a given `Source`, which is useful for applying transformations using akka-streams

Any linear graph that goes from a `Source` to a `Sink` via any combination of `Flow`s or whatever is strictly equivalent to a single `Source` -> `Sink`: that's why `runStream` works well for most use cases, even though our only input is a mutator of a `Source`.

However, we might want a non-linear graph: for example, there might be a "fork" where we want to complete some messages early as they require further processing. With `runStream` those messages would still have to be propagated all the way so that they can be sunk into an SQS ack at the end. 

This can sometimes be worked around by merging `Flow`s with `flatMapConcat` etc, but I think that it's cases like this where akka-streams can seem needlessly complex - because we don't have access to all of its features. This PR implements `runGraph` which gives us a `Source` of messages and a `Sink` to ack them, and leaves it up to us to put whatever graph we want between them.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: minor
+
+Adds the option to pass to SQSStream a function that creates an entire RunnableGraph from a source and sink.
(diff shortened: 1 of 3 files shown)


kenoir:
Can you explain briefly why `SQSStream` is awkward to use, and in what cases this change improves things?

kenoir (approved):

PR #75, opened 2021-02-15
Add randomSample to RandomGenerators

jamieparkinson:
This is a useful thing to have for taking members from an existing sequence:

```scala
val things = Seq(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
val randomThings = randomSample(things, size = 3) // e.g. Seq(7, 2, 9); 3 random elements of `things`
```

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Adds a method to RandomGenerators for taking a random sample from a given Seq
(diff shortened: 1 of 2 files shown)


jamesgorrie:
> for taking members from an existing sequence

Could you give an example?

jamieparkinson:
Added an example in the PR description

jamesgorrie (approved):

PR #77, opened 2021-03-11
Add elasticsearch and elasticsearch_typesafe from catalogue repo

alicefuzier:


RELEASE.md
```diff
@@ -0,0 +1,3 @@
+RELEASE_TYPE: minor
+
+Add `elastisearch` and `elasticsearch_typesafe` modules
```
alexwlchan, line 3:
```suggestion
Add `elasticsearch` and `elasticsearch_typesafe` modules
```

elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/model/CanonicalId.scala
```diff
@@ -0,0 +1,9 @@
+package uk.ac.wellcome.elasticsearch.model
+
+trait CanonicalId[T] {
+  def canonicalId(t: T): String
+}
```
alexwlchan, line 5: This is veering into catalogue terminology, but probably okay.

alexwlchan (approved):
I would shove this into the weco.elasticsearch namespace, but that's not a blocker. LGTM.

elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/model/CanonicalId.scala
alicefuzier: I think the naming of it comes from the catalogue, but really it's just a type class that gives an index id. So IMO the function is elastisearch specific but the name is historically tied to the catalogue because there we use canonicalIds as index ids. I could just rename as `IndexId`?

PR #79, opened 2021-03-21
Delete SQS messages in batches of 10, not individually

alexwlchan:
A quick glance at Cost Explorer revealed that we spend ~$100/mo just deleting messages. Literally just to tell SQS “yeah, we did that thing, you can forget about it now”.

🤯 

AWS have an easy fix for this – rather than deleting messages individually, you [use the DeleteMessageBatch action](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-batch-api-actions.html). This turns out to be a one-line change in the constructor we use for our Alpakka SQS sink, so I've just gone ahead and done it.

I don't expect it to make a big difference, but it won't be nothing either.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,4 @@
+RELEASE_TYPE: patch
+
+SQSStream and AlpakkaSQSWorker are now using DeleteMessageBatch rather than DeleteMessage.
+This should have no user-visible effects.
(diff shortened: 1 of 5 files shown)


kenoir (approved):
👏

PR #84, opened 2021-04-28
bump elastic version

jamesgorrie:
Bumping elastic to the version we actually use. 
Also give us some query parameters that I would like to use.


alexwlchan (approved):
You'll need to write a release note.

RELEASE.md

```
RELEASE_TYPE: [patch|minor|major]

Description of changes.
```

It gets compiled into the changelog when BuildKite cuts the release.

jamesgorrie:
@alexwlchan given it's a minor release change, would you consider this to be minor given we trust ES to follow semver definitions?

alexwlchan:
I think minor is fine, although we're not strict about semver in this repo tbh.

kenoir (approved):

PR #85, opened 2021-04-29
Add Tracing

kenoir:
Adds Tracing trait supporting APM tracing to typesafe_app lib, in order to make it available to all apps by default.

Will replace https://github.com/wellcomecollection/catalogue-api/blob/main/search/src/main/scala/uk/ac/wellcome/platform/api/search/Tracing.scala

### DIFF
--- a/typesafe_app/src/main/scala/uk/ac/wellcome/Tracing.scala
+++ b/typesafe_app/src/main/scala/uk/ac/wellcome/Tracing.scala
@@ -1,0 +1,172 @@
+package uk.ac.wellcome
+
+import java.util.concurrent.TimeUnit
+
+import akka.actor.ActorSystem
+import akka.dispatch._
+import co.elastic.apm.api.{ElasticApm, Transaction}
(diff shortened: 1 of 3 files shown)


typesafe_app/src/main/scala/uk/ac/wellcome/Tracing.scala
```diff
@@ -0,0 +1,172 @@
...
+    spanType: String = "",
+    subType: String = "",
+    action: String = "")(wrappedFunction: => Future[T]): Future[T] = {
+    val span = Tracing.currentTransaction
+      .startSpan(spanType, subType, action)
+      .setName(name)
+
+    wrappedFunction.transform { res =>
+      res match {
+        case Success(_)      =>
```
alexwlchan, line 96: This has been copied correctly from the original code; I'm just surprised that it works – I've never seen this before.

alexwlchan (approved):

kenoir:

PR #86, opened 2021-04-30
Add WellcomeHttpApp to http lib

kenoir:
This is duplicated between API services in storage & search - we should share it.

### DIFF
--- a/http/src/main/scala/weco/http/models/ErrorResponse.scala
+++ b/http/src/main/scala/weco/http/models/ErrorResponse.scala
@@ -1,0 +1,44 @@
+package weco.http.models
+
+import java.net.URL
+
+import akka.http.scaladsl.model.StatusCode
+import io.circe.generic.extras.JsonKey
+
+case class UserErrorResponse(
+  @JsonKey("@context") context: String,
+  httpStatus: Int,
+  description: String,
+  label: String,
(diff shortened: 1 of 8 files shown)


http/src/main/scala/weco/http/models/ErrorResponse.scala
```diff
@@ -0,0 +1,44 @@
...
+)
+
+case object InternalServerErrorResponse {
+  def apply(context: URL, statusCode: StatusCode): InternalServerErrorResponse =
+    InternalServerErrorResponse(
+      context = context.toString,
+      httpStatus = statusCode.intValue(),
+      label = statusCode.reason()
+    )
+}
```
alexwlchan, line 44: Is this an opportunity to consolidate some models? See https://github.com/wellcomecollection/scala-libs/blob/main/http/src/main/scala/weco/http/models/DisplayError.scala

alexwlchan (approved):

http/src/main/scala/weco/http/models/ErrorResponse.scala
kenoir: I think eventually yes. That might also happen by using the rejection & exception handlers in the search service.

kenoir:

PR #90, opened 2021-05-05
Don't require an ExecutionContext value here

alexwlchan:
Everywhere else we use the global EC, and it avoids resolution headaches elsewhere if we're consistent.


kenoir:
This seems like a good idea - but i'm thinking perhaps there are situations where we need to dictate the execution context though? 

I seem to remember there being some weirdness in the storage service with execution contexts in the http app not being shared with other things making tests fail. If nothing springs to mind - just merge this.

kenoir (approved):

PR #92, opened 2021-05-05
Standardise how we define context URLs

alexwlchan:
A first step towards https://github.com/wellcomecollection/platform/issues/5164

We have context URLs defined in a bunch of different ways: as a string, as a URL, as a variable called `context` or `contextURL` or `contextUrl`. All of this gets extra fiddly and dull for a thing we don't care that much about.

This PR establishes a standard for defining context URLs, and I'll clean up the other services to match.

### DIFF
--- a/http/src/main/scala/weco/http/HasContextUrl.scala
+++ b/http/src/main/scala/weco/http/HasContextUrl.scala
@@ -1,0 +1,7 @@
+package weco.http
+
+import java.net.URL
+
+trait HasContextUrl {
+  def contextUrl: URL
+}
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,9 @@
-RELEASE_TYPE: patch
+RELEASE_TYPE: minor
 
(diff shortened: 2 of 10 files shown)


kenoir (approved):

http/src/main/scala/weco/http/HasContextUrl.scala
```diff
@@ -0,0 +1,7 @@
+package weco.http
+
+import java.net.URL
+
+trait HasContextUrl {
+  def contextUrl: URL
```
jamieparkinson, line 6: I have slightly missed the boat here, but might we prefer to use akka-http's `Uri` class? I know it's a dependency but we're already, uh, dependent on it

alexwlchan: Is there a reason to pick one over the other? Does `Uri` get us some shiny new features?

kenoir: Why would we use one over the other?

PR #93, opened 2021-05-06
Allow setting contextUrl for assertIsDisplayError

kenoir:
This makes `assertIsDisplayError` work outside scala-libs.

### DIFF
--- a/http/src/test/scala/weco/http/fixtures/HttpFixtures.scala
+++ b/http/src/test/scala/weco/http/fixtures/HttpFixtures.scala
@@ -139,7 +139,7 @@ trait HttpFixtures extends Akka with ScalaFutures with Matchers
         jsonResponse,
         s"""
            |{
-           |  "@context": "${contextUrl}",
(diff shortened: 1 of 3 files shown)


alexwlchan (commented):
Can this be a `def` instead of a `val`? We have a couple of places in the API tests where it's a function of some other value (I think `ApiConfig`) that I suspect won't work with `val`.

http/src/test/scala/weco/http/fixtures/HttpFixtures.scala
```diff
@@ -135,7 +139,7 @@ trait HttpFixtures extends Akka with ScalaFutures with Matchers
         jsonResponse,
         s"""
            |{
-           |  "@context": "${ExampleApp.contextUrl}",
+           |  "@context": "${contextUrl}",
```
alexwlchan, line 142: nit:

```suggestion
           |  "@context": "$contextUrl",
```

alexwlchan (approved):

PR #95, opened 2021-05-07
Force people to supply a host/username/password

alexwlchan:
An Elasticsearch client isn't useful without those things, so hard fail on startup if they aren't supplied.

I learnt this while trying (and failing) to deploy https://github.com/wellcomecollection/catalogue-api/issues/8

### DIFF
--- a/elasticsearch_typesafe/src/main/scala/uk/ac/wellcome/elasticsearch/typesafe/ElasticBuilder.scala
+++ b/elasticsearch_typesafe/src/main/scala/uk/ac/wellcome/elasticsearch/typesafe/ElasticBuilder.scala
@@ -8,21 +8,16 @@ import uk.ac.wellcome.typesafe.config.builders.EnrichConfig._
 object ElasticBuilder {
   def buildElasticClient(config: Config,
                          namespace: String = ""): ElasticClient = {
-    val hostname = config
-      .getStringOption(s"es.$namespace.host")
-      .getOrElse("localhost")
+    val hostname = config.requireString(s"es.$namespace.host")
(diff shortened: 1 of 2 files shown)


elasticsearch_typesafe/src/main/scala/uk/ac/wellcome/elasticsearch/typesafe/ElasticBuilder.scala
```diff
@@ -8,21 +8,16 @@ import uk.ac.wellcome.typesafe.config.builders.EnrichConfig._
...
+    val username = config.requireString(s"es.$namespace.username")
+    val password = config.requireString(s"es.$namespace.password")
+
     val port = config
       .getIntOption(s"es.$namespace.port")
       .getOrElse(9200)
     val protocol = config
       .getStringOption(s"es.$namespace.protocol")
       .getOrElse("http")
-    val username = config
```
kenoir, line 20: In local development I don't believe these are required, will this require updates for consumers of this lib?

alexwlchan: It won't affect the tests – those call `ElasticClientBuilder.create` directly, rather than via Typesafe config. You would need it if you were running the Docker container – I don't do that, but maybe other people are?

kenoir: I think you are correct - it won't affect dev workflow.

kenoir (approved):

PR #98, opened 2021-05-13
Create a TypedString type for values that are more than strings, but barely

alexwlchan:
This is spun out of a comment on https://github.com/wellcomecollection/catalogue-pipeline/pull/1607#discussion_r631735484

Is this behaviour useful? Yes, I think so.

Should it be copied in a dozen different places? Definitely not.

### DIFF
--- a/json/src/main/scala/weco/json/TypedString.scala
+++ b/json/src/main/scala/weco/json/TypedString.scala
@@ -48,7 +48,7 @@ trait TypedString[T <: TypedString[_]] {
 
   override def equals(that: Any): Boolean =
     that match {
-      case that if that.getClass == this.getClass =>
+      case that if this.canEqual(that) =>
         this.underlying == that.asInstanceOf[T].underlying
       case _ => false
     }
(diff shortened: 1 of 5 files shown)


json/src/main/scala/weco/json/TypedString.scala
```diff
@@ -0,0 +1,67 @@
...
+trait TypedString[T <: TypedString[_]] {
+  val underlying: String
+
+  override def toString: String = underlying
+
+  def canEqual(a: Any): Boolean = a.getClass == this.getClass
+
+  override def equals(that: Any): Boolean =
+    that match {
+      case that if that.getClass == this.getClass =>
```
jamesgorrie, line 51: is `canEqual` what we should be using?

alexwlchan: Yup, might as well.

jamesgorrie (approved):
Smashing.

kenoir (commented):
👏

PR #110, opened 2021-06-16
Automatically open downstream PRs for new releases of scala-libs

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5200

This script runs at the end of a release build on main, in particular it:

* Clones the downstream repo
* Bumps the version in `Dependencies.scala`
* Commits that change and pushes it back to the repo
* Opens a PR for that change

### DIFF
--- a/.buildkite/scripts/open_downstream_prs.py
+++ b/.buildkite/scripts/open_downstream_prs.py
@@ -1,16 +1,52 @@
 #!/usr/bin/env python
 
+import contextlib
 import os
 import tempfile
 
 from commands import git
 
 
+@contextlib.contextmanager
+def working_directory(path):
+    """
+    Changes the working directory to the given path, then returns to the
+    original directory when done.
+    """
+    prev_cwd = os.getcwd()
+    os.chdir(path)
+    try:
+        yield
+    finally:
+        os.chdir(prev_cwd)
+
+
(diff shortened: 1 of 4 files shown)


.buildkite/scripts/open_downstream_prs.py
```diff
@@ -0,0 +1,144 @@
...
+    finally:
+        shutil.rmtree(repo_dir)
+
+
+def update_scala_libs_version(new_version):
+    old_lines = list(open("project/Dependencies.scala"))
+
+    with open("project/Dependencies.scala", "w") as out_file:
+        for line in old_lines:
+            if line.startswith("  val defaultVersion"):
```
kenoir, line 65: As we aren't parsing this file perhaps we should either use a comment in the downstream repos in `project/Dependencies.scala` to identify the line to update?

Something like:

```scala
val defaultVersion = "26.17.7" #LIBUPDATE: This line may be automatically update from the scala-libs build,  please keep this comment. 
```

Keep the version updating simple but make it a little more robust and add a placeholder to downstream repos so it will be clearer what is happening there?

alexwlchan: Good suggestion, done.

kenoir (approved):

PR #116, opened 2021-06-17
Make refresh_interval updatable

jamesgorrie:
Allows us to update the [`refresh_interval`](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-refresh-interval-setting) of an index.

The plan would be to set the `refresh_interval` to `-1` while reindexing, which is fine as we're generally not reindexing into and index that is a `search` index.


.bsp/sbt.json
```diff
@@ -1 +1 @@
-{"name":"sbt","version":"1.4.1","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/Users/jamesgorrie/Library/Java/JavaVirtualMachines/adopt-openj9-11.0.11/Contents/Home/bin/java","-Xms100m","-Xmx100m","-classpath","/Users/jamesgorrie/Library/Application Support/JetBrains/IntelliJIdea2021.1/plugins/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp"]}
\ No newline at end of file
+{"name":"sbt","version":"1.4.1","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/Users/jamesgorrie/.sdkman/candidates/java/11.0.10.j9-adpt/bin/java","-Xms100m","-Xmx100m","-classpath","/Users/jamesgorrie/Library/Application Support/JetBrains/IntelliJIdea2021.1/plugins/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp"]}
```
jamesgorrie, line 1: Pfft. Not sure what this is, will revert.

alexwlchan: I think we probably want to just exclude this file in `.gitignore`.

elasticsearch/docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 elasticsearch:
-  image: "docker.elastic.co/elasticsearch/elasticsearch:7.12.2"
+  image: "docker.elastic.co/elasticsearch/elasticsearch:7.12.1"
```
jamesgorrie, line 2: This was set to the scala4s version, there is no 7.12.2 in elasticsearch.

.bsp/sbt.json
jamesgorrie: Yip - removed.

kenoir (approved):

PR #117, opened 2021-06-22
convert index config to case class

jamesgorrie:
Better reflects how we use it as a structure with varying values of immutable data.

Adds ops to create an `IndexConfig` from typesafe `Config`.

ref: https://github.com/wellcomecollection/catalogue-pipeline/issues/1694

### DIFF
--- a/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
+++ b/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
@@ -22,14 +22,21 @@ object RefreshInterval {
   final case class On(duration: FiniteDuration) extends RefreshInterval
 }
 
-trait IndexConfig {
+trait IndexConfigTrait {
   def mapping: MappingDefinition
   def analysis: Analysis
   def shards: Int = 1
   def refreshInterval: RefreshInterval = RefreshInterval.Default
 }
 
-object NoStrictMapping extends IndexConfig {
+case class IndexConfig(mapping: MappingDefinition,
+                       analysis: Analysis,
+                       override val shards: Int = 1,
+                       override val refreshInterval: RefreshInterval =
+                         RefreshInterval.Default)
+    extends IndexConfigTrait
+
+object NoStrictMapping extends IndexConfigTrait {
   val analysis: Analysis = Analysis(analyzers = List())
   val mapping: MappingDefinition = MappingDefinition()
 }
--- a/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
+++ b/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
@@ -37,7 +37,7 @@ object IndexConfig {
 
   implicit class IndexConfigOps(indexConfig: IndexConfig) {
     def withRefreshIntervalFromConfig(config: Config): IndexConfig = {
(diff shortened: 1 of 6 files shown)


elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
```diff
@@ -22,14 +23,25 @@ object RefreshInterval {
...
+object IndexConfig {
+  def empty =
+    IndexConfig(
+      analysis = Analysis(analyzers = List()),
+      mapping = MappingDefinition())
 
-object NoStrictMapping extends IndexConfig {
-  val analysis: Analysis = Analysis(analyzers = List())
-  val mapping: MappingDefinition = MappingDefinition()
+  implicit class IndexConfigOps(indexConfig: IndexConfig) {
```
alexwlchan, line 38: Minor: we might want to put this in the ES typesafe lib, so it lives alongside all the other ES-related config. Or make it a boolean flag.

elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
```diff
@@ -22,14 +23,25 @@ object RefreshInterval {
...
+    def withRefreshIntervalFromConfig(config: Config): IndexConfig = {
+      val isReindexing = config.getBoolean(s"es.is_reindexing")
```
alexwlchan, line 40:
```suggestion
      val isReindexing = config.getBoolean("es.is_reindexing")
```

Wherever you put this, you don't need s-string interpolation.

elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
```diff
@@ -22,14 +23,25 @@ object RefreshInterval {
...
+                       analysis: Analysis,
+                       shards: Int = 1,
+                       refreshInterval: RefreshInterval =
+                         RefreshInterval.Default)
+
+object IndexConfig {
+  def empty =
+    IndexConfig(
+      analysis = Analysis(analyzers = List()),
+      mapping = MappingDefinition())
```
alexwlchan, line 36: Maybe we should just make these defaults on `IndexConfig`? There seems to be a lot of `analysis = Analysis(Nil)` further down.

alexwlchan (approved):

elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
jamesgorrie: There is a lot in testing, but I think the point is to not create empty / unstrict config in prod code.

elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/IndexConfig.scala
jamesgorrie: We could but it would be annoying to have to include things from two sources... or am I being a lazy for the wrong reasons here?

Currenty you don't need to import anything extra. If this was in typesafe libs
```
import weco.elastictypesafe.IndexConfigOps
class Main extends IndexConfigOps {}
```

jamesgorrie: Going to merge this, if we want to do this, it's another major bump, but no biggy.

PR #121, opened 2021-07-05
Remove the use of context URLs from our HTTP libraries

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5235


kenoir (approved):
As long as this is ok from @jtweed, let's do it!

PR #122, opened 2021-07-05
Use a Duration consistently for visibility timeout

alexwlchan:
I forgot to do this in https://github.com/wellcomecollection/scala-libs/pull/120


kenoir (approved):

alexwlchan:
I'm going to hold off on merging this because it will cause breakages in the automated bump that I can't deal with today, and I don't want to block merging of #123.

PR #123, opened 2021-07-08
Exit service if fatal error encountered

kenoir:
This release removes the unused type WellcomeApp and ensures that errors starting & configuring a service result in a service exiting.

Previously exceptions thrown inside `runWithConfig` would cause a service to stall rather than exit.


typesafe_app/src/main/scala/weco/typesafe/WellcomeTypesafeApp.scala
```diff
@@ -1,11 +1,31 @@
...
+
+  def runWithConfig(builder: Config => Runnable) =
+    try {
+      info(s"Starting service")
+
+      run {
+        builder(ConfigFactory.load)
+      }
+
+      info(s"Shutting down service")
```
alexwlchan, line 24: nit:

```suggestion
      info("Shutting down service")
```

typesafe_app/src/main/scala/weco/typesafe/WellcomeTypesafeApp.scala
```diff
@@ -1,11 +1,31 @@
...
+trait WellcomeTypesafeApp extends App with Logging {
+  protected def run(service: Runnable) =
+    Await.result(service.run(), Duration.Inf)
+
+  protected def exit(statusCode: Int = 0): Unit =
+    System.exit(statusCode)
+
+  def runWithConfig(builder: Config => Runnable) =
+    try {
+      info(s"Starting service")
```
alexwlchan, line 18: nit:

```suggestion
      info("Starting service")
```

alexwlchan (approved):

PR #125, opened 2021-07-14
Remove all the custom AWS config code; defer to env vars

alexwlchan:
The endpoint is the one place we might legitimately want to change that isn't exposed by env vars (e.g. if running in docker-compose), but that's a bit fiddly with the AWS v1 SDK we're using for S3 -- you need to supply the region at the same time.

This is a good reason to finally upgrade our S3 SDK to V2, but I don't want to do that right now.


kenoir (approved):

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/5241

PR #127, opened 2021-07-27
Fix a bug where partial locks weren't being released correctly

alexwlchan:
e.g. if you tried to lock (A, B, C), successfully locked A and B but failed to lock C, then the locks for A and B wouldn't be released. Now they get released correctly.

I suspect this is the true culprit behind the bug I was chasing in https://github.com/wellcomecollection/platform/issues/5211

The immediate motivation for this is fixing an issue in the storage service where bags get "stuck" if you try to ingest multiple bags with the same ingest ID in quick succession, but I suspect it may also be causing us issues in the matcher that will be fixed by this patch.

### DIFF
--- a/storage/src/test/scala/weco/storage/locking/LockingServiceTestCases.scala
+++ b/storage/src/test/scala/weco/storage/locking/LockingServiceTestCases.scala
@@ -25,7 +25,7 @@ trait LockingServiceTestCases[Ident, ContextId, LockDaoContext]
   val overlappingLockIds: Set[Ident] = commonLockIds ++ nonOverlappingLockIds
   val differentLockIds = Set(createIdent, createIdent, createIdent)
 
-  def withLockingService[R](testWith: TestWith[LockingServiceStub, R]): R =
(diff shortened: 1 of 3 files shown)


storage/src/test/scala/weco/storage/locking/LockingServiceTestCases.scala
```diff
@@ -188,5 +188,47 @@ trait LockingServiceTestCases[Ident, ContextId, LockDaoContext]
...
+    // prevented proc2 from proceeding when it was retried.
+    //
+    it("releases any already-acquired locks if one of the IDs fails to lock") {
+      val commonId = createIdent
+
+      val lockSet1 = Set(commonId, createIdent, createIdent)
+      val lockSet2 = Set(createIdent, commonId, createIdent)
+
+      println(s"lockSet1 = $lockSet1")
+      println(s"lockSet2 = $lockSet2")
```
alexwlchan, line 215: The printing here is intentional – if the lock dao is non-empty, you want to see which set failed to unlock correctly.

kenoir (approved):
Better than the other kind of locking failure!

PR #132, opened 2021-08-19
Pull SierraSource from catalogue-api stacks library

kenoir:
This change adds all the Sierra specific logic & models from the catalogue-api stacks library to reduce the common-lib in that project and keep same-domain code together.

### DIFF
--- a/sierra/src/main/scala/weco/sierra/http/SierraSource.scala
+++ b/sierra/src/main/scala/weco/sierra/http/SierraSource.scala
@@ -1,0 +1,155 @@
+package weco.sierra.http
+
+import akka.http.scaladsl.model.{HttpEntity, StatusCodes}
+import akka.http.scaladsl.model.Uri.Path
+import akka.http.scaladsl.unmarshalling.{Unmarshal, Unmarshaller}
+import akka.stream.Materializer
+import io.circe.Encoder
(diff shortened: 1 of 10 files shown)


sierra/src/main/scala/weco/sierra/http/SierraSource.scala
```diff
@@ -0,0 +1,155 @@
...
+import java.time.{Instant, ZoneId}
+import java.time.format.DateTimeFormatter
+
+import weco.http.client.{HttpClient, HttpGet, HttpPost}
+import weco.sierra.models.errors.{SierraErrorCode, SierraItemLookupError}
+import weco.sierra.models.fields.{SierraHoldRequest, SierraHoldsList, SierraItemDataEntries}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class SierraSource(client: HttpClient with HttpGet with HttpPost)(
```
alexwlchan, line 21: A thing I've been wondering about: is it worth splitting SierraSource into items and holds-specific logic?

Mostly because of the PII distinction – I think user holds count as PII, and items definitely don't – and a while back I wondered about splitting it so you knew if you were bringing in code to look up holds or not.

But not a big thing.

alexwlchan (approved):
Looks fine, modulo release note.

PR #133, opened 2021-08-24
Inspect queue size based on metrics, not waiting for messages to appear

alexwlchan:
You might have noticed that any test that relies on an SQS queue has gotten a lot slower recently – like, 30 seconds per test slower. That's because when you call `assertQueueSize`, we were waiting for 1.5x the queue visibility timeout to elapse before doing anything. This delay was mostly tolerable when the default timeout was 1s (wait 3s to inspect a queue and its DLQ), but I recently upped that timeout to 5s to avoid flaky tests in CI. So if you inspect two queues, you went from a 3 second to a 15 second wait. Oof.

This patch refactors the SQS queue fixture to avoid that delay, which should make all our tests much faster.

Thing is, waiting for visibility timeout to elapse isn't ideal. The theory went: _"I want to know how many messages are on a queue. I should wait for the visibility timeout to expire, then fetch from the queue and see what I get."_ This is fine… if you're the only person reading from the queue. But in all the tests where we create mock queues, we've got a worker that's also reading from the queue.

So we could wait for this timeout, fetch from the queue, and see the wrong size… because the worker got there before us. So we're already inspecting the MESSAGES_IN_FLIGHT metric to see if somebody else has some messages. This patch updates us to _only_ look at that metric, which is the most sensible way to assess queue size.

Plus some general tweaking around how we look up queue attributes, and making a few low-level methods private to encourage downstream code towards the helper methods.

### DIFF
--- a/messaging/src/test/scala/weco/messaging/fixtures/SQS.scala
+++ b/messaging/src/test/scala/weco/messaging/fixtures/SQS.scala
@@ -166,15 +166,12 @@ trait SQS extends Matchers with Logging with RandomGenerators {
     testWith(stream)
   }
 
-  def createNotificationMessageWith(body: String): NotificationMessage =
-    NotificationMessage(body = body)
-
   def createNotificationMessageWith[T](message: T)(
     implicit encoder: Encoder[T]): NotificationMessage =
-    createNotificationMessageWith(body = toJson(message).get)
+    NotificationMessage(body = toJson(message).get)
 
   def sendNotificationToSQS(queue: Queue, body: String): SendMessageResponse = {
-    val message = createNotificationMessageWith(body = body)
+    val message = NotificationMessage(body = body)
 
     sendSqsMessage(queue = queue, obj = message)
   }
--- a/messaging/src/test/scala/weco/messaging/fixtures/SQS.scala
+++ b/messaging/src/test/scala/weco/messaging/fixtures/SQS.scala
@@ -79,11 +79,6 @@ trait SQS extends Matchers with Logging with RandomGenerators {
       .toMap
   }
 
-  private def getQueueAttribute(
-    queue: Queue,
-    attributeName: QueueAttributeName): String =
-    getQueueAttributes(queue.url, attributeName)(attributeName)
-
   def createQueueName: String =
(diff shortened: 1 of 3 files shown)


messaging/src/test/scala/weco/messaging/fixtures/SQS.scala
```diff
@@ -204,58 +195,49 @@ trait SQS extends Matchers with Logging with RandomGenerators {
       .get
   }
 
-  def noMessagesAreWaitingIn(queue: Queue): Assertion = {
```
jamesgorrie, line 207: Removing this method would have knock on effects here, which might make this a major change?

https://github.com/search?q=org%3Awellcomecollection+noMessagesAreWaitingIn&type=code

alexwlchan: This is an interesting one, and depends on how you interpret the versioning. It looks SemVer-ish, but we've never been particularly consistent about that.

Because it's just for us, I tend to use the major/minor/patch versions as an indication of how much work is involved:

* patch = should be a drop-in replacement
* minor = maybe some small tweaking involved, but nothing substantial
* major = fundamental change to how the library works, expect downstream to have big patches

In the external uses of `noMessagesAreWaitingIn`, we should really be using `assertQueueEmpty`. That's only a small change when I do the downstream PRs, so I think of this as a minor patch.

jamesgorrie: Cool cool. Why not just use semver definitions though?

jamesgorrie (approved):

messaging/src/test/scala/weco/messaging/fixtures/SQS.scala
alexwlchan: By a strict interpretation of SemVer, pretty much every change to scala-libs is breaking in some way (e.g. removing a method that's totally unused across the platform is a SemVer major version bump, even though it's a no-op for us), at which point it's not very helpful.

TBH I'd be tempted to switch to something more like monotonic integers, but never quite got round to it.

PR #139, opened 2021-09-08
Remove a couple of unused imports from tests

alexwlchan:


kenoir (approved):
Is it just in tests that the linter is OK with unused imports?

alexwlchan:
I don't think it's okay with unused imports, I think I misspoke – these are *duplicate* imports.

So it'd warn for:

```scala
import weco.storage.S3TypedStore

# EOF
```

but not for:

```scala
import weco.storage.S3TypedStore
import weco.storage.S3TypedStore

val store = S3TypedStore[T]()
# EOF
```

PR #141, opened 2021-09-14
Always use compression when talking to Elasticsearch

alexwlchan:
We wrote all this code for compressing our Elasticsearch traffic… why not use it?!

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+RELEASE_TYPE: minor
+
+This release changes our Elasticsearch client to always use compression.
+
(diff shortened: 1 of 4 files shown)


jamieparkinson (commented):
Would it perhaps make more sense to keep the option but to change the default?

alexwlchan:
> Would it perhaps make more sense to keep the option but to change the default?

I considered it, but is there any scenario in which we wouldn't want to use compression?

jamieparkinson:
iirc we saw that compression increases CPU load on both the cluster and the application using the client, we might want to be able to turn it off. Not especially bothered though.

jamieparkinson (approved):

PR #145, opened 2021-09-16
Update SierraSource.lookupItemEntries to request more fields

kenoir:
In order to ensure that we have consistent item data for Sierra items between the items API and the search API we need to make sure the same data is sourced from Sierra in both cases.

This change adds the "status" and "varFields" fields to those requested when we look up item data directly from the sierra API.

This change resolves an issue where we were not seeing notes fields on items from the Items API (https://github.com/wellcomecollection/platform/issues/5305)


alexwlchan (approved):
This looks fine, although I wonder if we'd be better off making the list of fields a parameter you pass in, so we can change it up as we need?

kenoir:
>This looks fine, although I wonder if we'd be better off making the list of fields a parameter you pass in, so we can change it up as we need?

I think this makes sense and this change can function as a step on the way. We might want to create an ADT around the params, rather than allowing strings to be handed through.

PR #147, opened 2021-10-28
Remove the unused DockerCompose settings

alexwlchan:
We get a warning that they're unused whenever we run sbt, so we might as well remove them (as we've done in other repos).


alexwlchan:
Closing for now; it's not worth investigating why the Elasticsearch tests are unhappy.

PR #150, opened 2021-11-05
Add implicit decoders/encoders for NotificationMessage

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5359

Although it's a very small class, we use it *everywhere*. Let's shove an implicit encoder/decoder on the companion object (so we don't need to import anything extra), and save the compiler some work.


jamesgorrie (approved):
Lovely 🚤

PR #158, opened 2022-01-21
Add retrying logic for transient AWS network errors

alexwlchan:
The release note explains the broad strokes.

The size of the diff is because I moved `RetryableError` into `typesafe_app` (which is really becoming a sort of "core" library), so we'd have a single definition rather than multiple ones flying around. 

For https://github.com/wellcomecollection/platform/issues/5419


jamieparkinson (approved):
Nice, I like the clear error messages

PR #160, opened 2022-01-21
Remove a lot of unnecessary traits and classes in the messaging lib

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5419

The *actual* issue is what I think is a design flaw in AlpakkaSQSWorker (or one of its parent traits), which I'll highlight in the diff below – but there's so much abstraction in the current design that it's very difficult to spot.

This patch collapses a lot of the abstractions into concrete classes, and does away with a lot of generic traits that could be extended but in practice never were. I've also tried to get away from some terms that are overloaded in the platform (e.g. "work", "transform") in favour of alternatives.

The key class is `Worker`, from which everything else falls out.

We still have two ways to read messages from SQS (AlpakkaSQSWorker in the storage service, SQSStream in the catalogue pipeline). Ideally we'd settle on a single approach, but that's a lot more work than I want to do now.

This diff is too big to review sensibly, so I'm going to break it down into smaller PRs:

- https://github.com/wellcomecollection/scala-libs/pull/161


messaging/src/main/scala/weco/messaging/worker/Processor.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  private def process(
+    workEither: Either[Throwable, Input]): Future[Result[Summary]] =
+    workEither.fold(
+      e => Future.successful(TerminalFailure[Summary](e)),
+      w =>
+        doProcessing(w) recover {
+          case e => TerminalFailure[Summary](e)
+      }
+    )
```
alexwlchan, line 75: These are the interesting lines of code.

The output of `doProcessing` is `Future[Result[_]]`, where `Result` can be success/failure/retryable failure.

This function should *never* return a failed future. If something goes wrong in the message processing, the caller should return a successful future with a failure result inside, e.g.

```scala
def doProcessing(key: DynamoKey) {
  dynamoClient.getItem(key) match {
    case Found(_)    => Future.successful(Success(…))
    case NotFound(_) => Future.successful(TerminalFailure(…))
  }
}
```

If this function returns a failed future, it means there's an exception somewhere that we're not handling – and then this line deletes it from the SQS queue.

Writing it out, I remember why we did it this way – in the storage service, if something goes wrong unexpectedly, we want to bail out immediately, rather than push on and potentially make things worse. So removing it from the main queue is correct, but we should put it on the DLQ as well.

So here's my suggestion (which I'll implement separately, so it can be sensibly reviewed):

We introduce a new failure type `UnhandledFailure`, which is only returned in this branch. If that failure is return, we apply a new `unhandledFailureAction`, which move the message straight to the DLQ. Benefits of this approach:

- We get the message off the main queue after the first unhandled failure
- The message is still on a DLQ, so we can redrive it manually if we think it's a transient failure
- It gets flagged by our regular DLQ alarms, so we'll see something has gone wrong (rather than waiting a day and a half to see the ingest has failed, as we did today)

And possibly we'd mix in some logic here to identify transient failures automatically, and mark them as RetryableFailure instead (e.g. the AWS DNS errors we saw).

alexwlchan:
Closing as superseded by other PRs.

PR #162, opened 2022-01-21
Tidy up the MonitoringProcessor trait and related pieces

alexwlchan:
For wellcomecollection/platform#5419.

In particular, we have this trait MonitoringProcessor in #160, which has two functions:

```
trait MonitoringProcessor {
  def recordStart()
  def recordEnd()
}
```

There were two implementations (MetricsMonitoringProcessor and OpenTracingMonitoringProcessor), but we binned the latter in #160, and the `recordStart` in the remaining implementation was `Future.successful(Right(Instant.now))` – which is much cleaner to inline and call `startTime` where that method is called. And once you tug on that thread, you realise there's not much point in keeping the trait around, and it's better inlined into `Worker`.


messaging/src/main/scala/weco/messaging/sqsworker/alpakka/AlpakkaSQSWorker.scala
```diff
@@ -59,28 +55,3 @@ class AlpakkaSQSWorker[Work,
   val completedAction: SQSAction = (message: SQSMessage) =>
     MessageAction.delete(message)
 }
-
-object AlpakkaSQSWorker {
-  def apply[Work,
```
alexwlchan, line 64: All this `apply()` method got us was the ability to do

```scala
AlpakkaSQSWorker(config, monitoringProcessBuilder, process)
```

instead of

```scala
AlpakkaSQSWorker(config, monitoringProcessBuilder)(process)
```

I binned it rather than fix it, because I don't think that's especially useful. It might need us to tweak a few constructors downstream.

messaging/src/test/scala/weco/messaging/worker/WorkerTest.scala
```diff
@@ -1,189 +1,184 @@
...
     extends AnyFunSpec
     with Matchers
-    with Akka
     with ScalaFutures
-    with IntegrationPatience
     with WorkerFixtures
     with MetricsFixtures {
 
   it("successfully processes a work and increments success metrics") {
-    withMetricsMonitoringProcessor[MyWork, Unit]() {
```
alexwlchan, line 23: Although this test has a large diff, it's mostly de-denting where we used to have a `withMetricsMonitoringProcessor` fixture in favour of:

```scala
val namespace = s"ns-${randomAlphanumeric()}"
implicit val metrics = new MemoryMetrics()
```

messaging/src/main/scala/weco/messaging/worker/Worker.scala
```diff
@@ -42,29 +33,23 @@ trait Worker[Message,
...
-      localContext <- monitoringProcessor.recordStart(workEither, rootContext)
-      summary <- process(workEither)
-      _ <- log(summary)
-      _ <- monitoringProcessor.recordEnd(localContext, summary)
-    } yield
-      WorkCompletion(
-        message,
-        summary
-      )
+      (workEither, _) <- Future.successful(callTransform(message))
```
alexwlchan, line 48: You'll notice the second half of this tuple is now unused – I'm going to clean this up in a subsequent patch.

jamieparkinson (approved):
v helpful explanations in the comments, looks good!

PR #164, opened 2022-01-24
Simplify and inline the MessageTransform trait

alexwlchan:
For wellcomecollection/platform#5419, follows #162 and #163, pulled out of #160

In particular, in #162, we stopped using the second item of the 2-tuple returned by `MessageTransform.callTransform()` (https://github.com/wellcomecollection/scala-libs/pull/162#discussion_r790453669) – which returned a generic context that we no longer use. So this patch:

* Modifies that method to return a value, not a 2-tuple
* Deletes all the type parameters we no longer need
* Inlines the MessageTransform trait into Worker
* Renames the method to parseMessage rather than transform, which is an overloaded term in the platform


jamieparkinson (approved):
✨

PR #165, opened 2022-01-24
Simplify and inline the MessageProcessor trait

alexwlchan:
For wellcomecollection/platform#5419.

The `parseMessage` method in #164 is decidedly weird: it returns an `Either[Throwable, Work]`, but then we wrap it in another `Try` block to catch and wrap any exception that wasn't in the Left arm of the Either. This is unnecessarily defensive, and makes the code more verbose.

Let's change `parseMessage` to return a `Try[Work]`, then inline some of the logic that uses that result in the Worker trait. It also (finally) means we're getting to the code that caused the original issue – see the comment in the `Worker` trait.


messaging/src/main/scala/weco/messaging/worker/Worker.scala
```diff
@@ -39,21 +38,27 @@ trait Worker[Message, Work, Summary, Action]
...
     val startTime = Instant.now()
-    val workEither = doParseMessage(message)
 
     for {
-      result <- process(workEither)
+      // TODO: DeterministicFailure is the wrong choice here -- it will cause these
+      // messages to be marked as "completed", and deleted from the queue.
+      //
+      // In either case (we can't parse the message, or an unhandled exception in doWork),
+      // we should put the messages on a DLQ for further investigation.
```
alexwlchan, line 47: The suggested outcome (we should put the messages on a DLQ) is correct, but changing the error type isn't the way to do it – I'll return to this in a separate patch, but I wanted to leave a note here for now.

jamieparkinson (approved):

PR #168, opened 2022-01-24
Log a warning whenever we delete messages with a DeterministicFailure; rename result types

alexwlchan:
For wellcomecollection/platform#5419; follows #167.

This patch improves the documentation:

* A README for the messaging library, which explains the key components and why we have two classes (SQSStream and AlpakkaSQSWorker) that seem to do the same thing
* Better names for the `Result` classes used by AlpakkaSQSWorker, and brief comments explaining when to use them

And it changes the behaviour for failed messages: when an SQS message fails processing in AlpakkaSQSWorker, we:

1. Log a warning with the complete message (including the body)
2. Remove the message from the original queue

Previously we would just do (1), which was causing messages to disappear without trace.

I think there's more to do, in particular tweaking some of the retry logic, but I'll leave that for a separate patch. Just wrapping my head around this simple change has taken long enough. 😅

### DIFF
--- a/messaging/src/main/scala/weco/messaging/worker/models/Result.scala
+++ b/messaging/src/main/scala/weco/messaging/worker/models/Result.scala
@@ -10,7 +10,7 @@ sealed trait Result[Summary] {
     s"$name: ${summary.getOrElse("<no-summary>")}"
 }
 
-// An operation that failed, and can't be retried.
+// An operation that failed, and which the caller thinks can't be retried.
 //
 // e.g. trying to write an object to a non-existent S3 bucket.  This will always
 // fail, no matter how often we try.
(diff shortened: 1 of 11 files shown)


messaging/src/main/scala/weco/messaging/worker/models/Result.scala
```diff
@@ -10,21 +10,25 @@ sealed trait Result[Summary] {
     s"$name: ${summary.getOrElse("<no-summary>")}"
 }
 
-case class DeterministicFailure[Summary](
+// An operation that failed, and can't be retried.
+//
+// e.g. trying to write an object to a non-existent S3 bucket.  This will always
+// fail, no matter how often we try.
+case class TerminalFailure[Summary](
```
alexwlchan, line 17: I don't love the word "terminal" here, because in some cases we will retry these – but it feels like an improvement from "deterministic/non-deterministic".

jamieparkinson (approved):
Very nice, fan of the improved naming - have never been particularly comfortable with the previous naming

PR #175, opened 2022-03-07
Check a patron record for patron type

melanierogan:
We will eventually want to check if a patron is self registered, so we can check if they are allowed to request items from the collection. In order to check this, we will need to know the patron type. This PR adds a `lookupPatronType` method that will return a patron's `PatronType` if there is one, and `None` if there isn't. 

The work to improve error messaging for patron types that can't request is from this ticket, but a few more PRs are needed to finish this: https://github.com/wellcomecollection/wellcomecollection.org/issues/7645

My very first time writing Scala!


jamieparkinson (approved):
:tada:

PR #176, opened 2022-03-29
Allow access to notes field, make neededBy optional

jamieparkinson:
As I've alluded to in a comment, I started to pull a thread related to Circe encoders and Unmarshallers; we use a mix of https://github.com/hseeberger/akka-http-json and our own `CirceMarshalling` utility, which makes behaviour inconsistent when using some Circe features (namely, `Printer`s). This should probably be addressed at some point, but it looked like quite a long and painful thread so I ducked out in the end...


melanierogan (approved):
Makes sense to me, though you may want a more experienced set of scala eyes on this :)

PR #190, opened 2022-08-25
Trying to sort out shapeless/scala version chaos

jamieparkinson:
I have confirmed that setting this `scala-reflect` version on an application itself fixes the problem we've been seeing - I hope that setting it here (rather than having to use it in every application) will work too!


melanierogan (approved):
Sounds sensible, fingers crossed

PR #192, opened 2022-11-04
Remove the autogenerated sbt metadata files

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5626; mirrors https://github.com/wellcomecollection/catalogue-api/pull/568

It looks like we already ditched a bunch of the clever build logic here, let's go further and strip out the autogen metadata.


alexwlchan:
Replaced by #193.

PR #194, opened 2022-11-22
Remove the refresh interval settings from scala-libs

alexwlchan:
As discussed in https://github.com/wellcomecollection/catalogue-pipeline/issues/2238


paul-butcher (commented):
LGTM.  Over a hundred lines of debugging!

I plan to run a new catalogue reindex on Monday (for the LSH business), which I might like to run with this switched off, just to be sure, before we completely remove it.

That said, I suppose we do have more than just the one measurement, as you ran it successfully on Works a while ago

paul-butcher (approved):

alexwlchan:
> I plan to run a new catalogue reindex on Monday (for the LSH business), which I might like to run with this switched off, just to be sure, before we completely remove it.

Sure, I can hold this, there's no big rush.

PR #199, opened 2023-01-05
Preparing the way for the AWS Java V2 SDK for S3, part 1 (remove batchSize; small clean-ups)

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4785

Trying to migrate all of scala-libs in one go is too big to be manageable, so I'm going to break it down into smaller PRs. This has the first bit of refactoring for the V2 SDK:

* Remove some duplicate imports
* Construct requests as standalone variables. The V2 SDK changes how requests are constructed quite dramatically, e.g.

    ```scala
    new GetObjectRequest("my-bukkit", "cat.jpg")
    ```
 
    becomes

    ```scala
    GetObjectRequest.builder()
      .bucket("my-bukkit")
      .key("cat.jpg")
      .build()
    ```

    The latter has more clarity, but you really want to construct it as a separate variable before calling the SDK method.

    Pulling out those requests as separate variables now reduces the diff churn before the V2 upgrade. Compare:

    ```diff
    -s3Client.getObject(new GetObjectRequest("my-bukkit", "cat.jpg"))
    +val request = GetObjectRequest.builder()
    +  .bucket("my-bukkit")
    +  .key("cat.jpg")
    +  .build()
    +s3Client.getObject(request)
    ```

    and
    ```diff
    -val request = new GetObjectRequest("my-bukkit", "cat.jpg")
    +val request = GetObjectRequest.builder()
    +  .bucket("my-bukkit")
    +  .key("cat.jpg")
    +  .build()
     s3Client.getObject(request)
    ```

* Remove the `batchSize` parameter from S3ObjectLocationListing. It's not obvious if/how this can be achieved in the V2 SDK, and we never set it to a non-default value in production code, just in tests.  Let's get rid of it to simplify things.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,7 @@
+RELEASE_TYPE: minor
+
+Remove batchSize from S3ObjectLocationListing.
+
+This is never set to a non-default value, so there should be no impact on downstream code.
+
+Also contains some refactoring ahead of upgrading to the AWS Java V2 SDK for S3.
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -5,10 +5,10 @@ object Dependencies {
     val elasticApm = "1.22.0"
     val elastic4s = "8.3.2"
 
-    val aws = "1.11.504"
+    val aws1 = "1.11.504"
 
     // Moving what we can to version 2 of the AWS SDKs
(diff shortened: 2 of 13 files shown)


jamieparkinson (approved):
mmmm, Java APIs

PR #201, opened 2023-01-05
Preparing the way for the AWS Java V2 SDK for S3, part 3 (Remove the checkForExisting flag from the Transfer classes)

alexwlchan:
In practice, we always set `checkForExisting=true`, so let's remove the unsafe path where we might overwrite an object without checking first.

Closes https://github.com/wellcomecollection/platform/issues/5647, for https://github.com/wellcomecollection/platform/issues/4785

### DIFF
--- a/storage/src/main/scala/weco/storage/transfer/azure/AzureTransfer.scala
+++ b/storage/src/main/scala/weco/storage/transfer/azure/AzureTransfer.scala
@@ -191,16 +191,21 @@ trait AzureTransfer[Context]
       case Failure(err) => throw err
     }
 
-  override def transfer(src: SourceS3Object, dst: AzureBlobLocation): TransferEither =
+  override def transfer(src: SourceS3Object,
+                        dst: AzureBlobLocation): TransferEither =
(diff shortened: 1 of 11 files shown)


storage/src/main/scala/weco/storage/transfer/azure/AzureTransfer.scala
```diff
@@ -172,7 +170,7 @@ trait AzureTransfer[Context]
           }
       }
 
-      blockClient.commitBlockList(identifiers.toList.asJava, allowOverwrites)
+      blockClient.commitBlockList(identifiers.toList.asJava)
```
alexwlchan, line 172: When I first modified this line, I got the boolean the wrong way round; this method is an alias for:

```
commitBlockList(base64BlockIds, overwrite: false);
```

but now we get the default-safe behaviour of the Azure SDK without mucking around with our own booleans.

alexwlchan:
As a sidebar, I'd really like to get these classes pulled out of scala-libs and into the storage service.

jamieparkinson (approved):

PR #202, opened 2023-01-05
Use the AWS Java V2 SDK for S3, part 1 (use the V2 SDK in tests)

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/4785

I don't want to upgrade the library code and the tests at once, because there's a good chance I'll make the same mistake on both sides and miss it. 🙈 

So I'm going to update them in bits. 

This PR updates a bunch of the test code – we make assertions about the state of the S3 container using the V2 SDK – but without changing any of the library code it's testing. I'd like to get this merged and used everywhere as an initial sense check of the V2 SDK, then when I change the library code I'll be testing on slightly more solid ground.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,3 @@
+RELEASE_TYPE: patch
+
+Start using the AWS Java V2 SDK for S3 for testing S3-related code.  This doesn't change any library code.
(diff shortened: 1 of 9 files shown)


jamieparkinson (approved):
Feels rather odd that we can change tests independently of library code doesn't it

PR #207, opened 2023-01-06
Use the AWS Java V2 SDK for S3

alexwlchan:
This does the bulk of the work for https://github.com/wellcomecollection/platform/issues/4785

I'll need to fix some stuff downstream in the app repositories, but I think those will be pretty minor.

Each commit updates a single class, if that's easier to review.


storage/src/main/scala/weco/storage/store/s3/S3StreamWritable.scala
```diff
@@ -1,73 +1,57 @@
...
-  // For multipart PUT requests the size must be larger than each PART.
-  //
-  // To prevent this exception a constant maximum size is set
-  // assuming a maximum file of 600GB PUT as 10,000 multipart requests
-  // = 60MB ~ 100MB read limit
-  // this is a generous estimate and should be sufficient,
-  // also given x10 concurrent streams = 10x100MB = 1GB memory overhead which we are comfortable with.
-  // This change was tested to reproduce the error with a proxy that dropped traffic to simulate S3 network failure.
-  private val MB: Int = 1024 * 1024
-  private val BUFFER_READ_LIMIT: Int = 100 * MB
```
alexwlchan, line 43: I couldn't work out how to set the buffer read limit with the V2 SDK. 🤷 

The whole AsyncRequestBody stuff is a little bit mysterious and I think we might end up tweaking it if/when we see errors in production.

storage/src/main/scala/weco/storage/store/s3/S3StreamWritable.scala
```diff
@@ -76,28 +60,32 @@ trait S3StreamWritable
...
+        .upload(uploadRequest)
+        .completionFuture()
+        .join()
     } match {
-      case Success(_)   => Right(Identified(location, inputStream))
+      case Success(_) if inputStream.available() > 0 =>
+        Left(
+          IncorrectStreamLengthError(
+            new RuntimeException(
+              "Data read has a different length than the expected")))
```
alexwlchan, line 77: This is something that used to be handled for us by the V1 SDK – the problem is if the InputStreamWithLength specifies a length that's too short, and some of the bytes are left unread.

e.g.

```
InputStreamWithLength(bytes = "Hello world", length = 5)
```

I think passing a length to `AsyncRequestBody` means it's not going to read beyond the end of the stream, and doesn't notice the extra bytes sitting there. The upload will still succeed, but this is caught by our tests so I had to put this in.

storage/src/main/scala/weco/storage/store/s3/S3StreamWritable.scala
```diff
@@ -109,6 +97,15 @@ trait S3StreamWritable
...
           if exc.getMessage.startsWith("More data read than expected") =>
         IncorrectStreamLengthError(exc)
+
+      // e.g. Request content was only 1024 bytes, but the specified content-length was 1025 bytes.
+      case exc: SdkClientException
+          if exc.getCause
+            .isInstanceOf[IllegalStateException] && exc.getCause.getMessage
+            .contains("Request content was only") && exc.getCause.getMessage
+            .contains("but the specified content-length was") =>
+        IncorrectStreamLengthError(exc)
```
alexwlchan, line 107: Again dealing with content-length and making sure we map it to the right type; some of the error messages have changed between V1 and V2.

storage/src/main/scala/weco/storage/transfer/s3/S3Transfer.scala
```diff
@@ -100,23 +105,46 @@ class S3Transfer(transferManager: TransferManager, s3Readable: S3StreamReadable)
       result <- tryCopyToDestination(src, dst, transfer)
         .retry(maxAttempts = 3)
         .left
-        .map(err => TransferDestinationFailure(src, dst, err.e))
+        .map {
+          // Handle theb case where the source object doesn't exist
+          case err if err.e.isInstanceOf[NoSuchKeyException] =>
+            TransferSourceFailure(src, dst, err.e)
```
alexwlchan, line 111: in V1 this would have thrown inside `tryCopyFromSource()` when we call `transferManager.copy(copyRequest)`, but that's not the case in V2

storage/src/main/scala/weco/storage/transfer/s3/S3Transfer.scala
```diff
@@ -100,23 +105,46 @@ class S3Transfer(transferManager: TransferManager, s3Readable: S3StreamReadable)
...
+        .copyObjectRequest(copyObjectRequest)
+        .build()
 
     Try {
-      // This code will throw if the source object doesn't exist.
       transferManager.copy(copyRequest)
     } match {
       case Success(request) => Right(request)
+      case Failure(err: CompletionException) =>
+        Left(S3Errors.readErrors(err.getCause))
```
alexwlchan, line 147: These various CompletionException handlers are sprinkled around because that's what the transfer manager futures wrap the exception in.

storage/src/test/scala/weco/storage/listing/ListingTestCases.scala
```diff
@@ -27,7 +27,7 @@ trait ListingTestCases[Ident,
   def createPrefixMatching(id: Ident): Prefix
 
   def assertResultCorrect(result: Iterable[ListingResult],
-                          entries: Seq[Ident]): Assertion
+                          entries: Seq[Ident])(implicit context: ListingContext): Assertion
```
alexwlchan, line 30: This is used in the S3 listing tests; there `ListingResult` is `S3Object`, which doesn't know about the bucket, so we pass that in the context.

storage/src/test/scala/weco/storage/listing/s3/S3ObjectListingTest.scala
```diff
@@ -0,0 +1,23 @@
+package weco.storage.listing.s3
+
+import org.scalatest.Assertion
+import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.s3.model.S3Object
+import weco.storage.fixtures.S3Fixtures.Bucket
+import weco.storage.s3.S3ObjectLocation
+
+class S3ObjectListingTest extends S3ListingTestCases[S3Object] {
```
alexwlchan, line 9: This was S3ObjectSummaryListingTest; I changed the name of the class because S3Object (V2) has replaced S3ObjectSummary (V1).

storage/src/test/scala/weco/storage/services/s3/S3LargeStreamReaderTest.scala
```diff
@@ -48,33 +40,4 @@ class S3LargeStreamReaderTest
...
-    val bufferSize = 500
-
-    withLocalS3Bucket { bucket =>
-      val location = createS3ObjectLocationWith(bucket)
-      putStream(
-        location,
-        inputStream = createInputStream(length = bufferSize * 2)
-      )
-
-      val spyClient = Mockito.spy(s3Client)
```
alexwlchan, line 62: Mockito is unhappy with spying on the new S3Client class, because it's a "final class".

This test isn't testing any functionality we care about; it's testing an implementation detail, so I think it's safe to remove.

storage/src/test/scala/weco/storage/services/s3/S3SizeFinderTest.scala
```diff
@@ -42,42 +38,6 @@ class S3SizeFinderTest
...
-
-  it("finds the size of an object in Glacier") {
-    // Note: CloudServer (the Docker image we use for mocking S3 in tests) doesn't
-    // handle objects in Glacier correctly.
-    //
-    // Normally, calling GetObject on something in Glacier will return an error
-    // (InvalidObjectState), but GetObjectMetadata works fine.  We use spy() here
-    // because we can't rely on CloudServer to warn us itself.
-    //
-    // See https://github.com/scality/cloudserver/issues/2977
```
alexwlchan, line 56: This test has always been a faff, and again we're bitten by Mockito spying. I think we can remove it until we see it's an issue.

storage/src/main/scala/weco/storage/services/s3/S3SizeFinder.scala
```diff
@@ -1,48 +1,63 @@
...
+          GetObjectRequest
+            .builder()
+            .bucket(location.bucket)
+            .key(location.key)
+            .build()
 
         val s3Object = s3Client.getObject(getRequest)
-        val metadata = s3Object.getObjectMetadata
-        val contentLength = metadata.getContentLength
+        val contentLength = s3Object.response().contentLength()
```
paul-butcher, line 50: This replicates the existing code, but is it meaningful to get content length here, when the main reason for running this branch is to get a better error message when we already know it isn't going to work?

paul-butcher: I suppose the question is whether, in practice, this line is ever reached, or whether getObject is expected to always raise an exception. 

If there is an intermittent connectivity or timing problem, which I assume is expected from the fact this is run with retries, this does mean we essentially get two attempts at fetching the size for every retry.

paul-butcher (approved):

storage/src/main/scala/weco/storage/services/s3/S3SizeFinder.scala
alexwlchan: Good question – I'm not sure.

I have a vague memory of permissions that allow GetObject without HeadObject, but I'm not sure that makes any sense. I'm going to add a TODO note about this, then merge it.

PR #208, opened 2023-01-11
Do the multipart upload ourselves, not with TransferManager

alexwlchan:
There's a test in https://github.com/wellcomecollection/storage-service/pull/1046 that hangs forever. We have a collection of input streams in a sequence:

```mermaid
graph TD
   S3[InputStream from truncated_header.tar.gz] --> U[gzip-uncompressed InputStream of truncated_header.tar]
   U --> B[InputStream from individual tar entry 9.bin]
   B --> Up[being uploaded to S3 through TransferManager]
```

but the individual tar entry `9.bin` is truncated. This should be throwing an EOF exception, but for some reason the S3 client is hanging, waiting forever for more bytes that will never come. I think the exception is being swallowed somewhere inside the S3 TransferManager concurrency model, which uses a Java ExecutorService and I don't really understand properly.

I think we can skip using S3TransferManager, and do the multipart upload ourselves. 99% of the "new" code in this PR comes from the snapshot generator, where we already do this – the only difference is that we don't know the size of a snapshot in advance; we have to just keep creating new parts as we need them. Here, we know how big the object should be in advance, so we know up front how many parts we'll need.

This passes with all the existing S3 tests in scala-libs, and I've confirmed it works in the failing storage service test also.

Here's the corresponding code in the snapshot generator: https://github.com/wellcomecollection/catalogue-api/blob/main/snapshots/snapshot_generator/src/main/scala/weco/api/snapshot_generator/storage/S3Uploader.scala

### DIFF
--- a/storage/src/main/scala/weco/storage/s3/S3Errors.scala
+++ b/storage/src/main/scala/weco/storage/s3/S3Errors.scala
@@ -46,6 +46,10 @@ object S3Errors {
     case exc: S3Exception if exc.statusCode() == 500 =>
       new StoreWriteError(exc) with RetryableError
 
+    // e.g. S3Exception: Object key is too long. Maximum number of bytes allowed in keys is 915.
+    case exc: S3Exception if exc.getMessage.startsWith("Object key is too long") =>
+      InvalidIdentifierFailure(exc)
+
     case exc => StoreWriteError(exc)
   }
 }
--- a/storage/src/main/scala/weco/storage/s3/S3Errors.scala
+++ b/storage/src/main/scala/weco/storage/s3/S3Errors.scala
@@ -47,7 +47,8 @@ object S3Errors {
       new StoreWriteError(exc) with RetryableError
 
     // e.g. S3Exception: Object key is too long. Maximum number of bytes allowed in keys is 915.
-    case exc: S3Exception if exc.getMessage.startsWith("Object key is too long") =>
+    case exc: S3Exception
+        if exc.getMessage.startsWith("Object key is too long") =>
       InvalidIdentifierFailure(exc)
 
     case exc => StoreWriteError(exc)
--- a/storage/src/main/scala/weco/storage/store/s3/S3MultipartUploader.scala
+++ b/storage/src/main/scala/weco/storage/store/s3/S3MultipartUploader.scala
@@ -3,7 +3,13 @@ package weco.storage.store.s3
 import grizzled.slf4j.Logging
 import software.amazon.awssdk.core.sync.RequestBody
 import software.amazon.awssdk.services.s3.S3Client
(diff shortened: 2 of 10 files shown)


storage/src/main/scala/weco/storage/s3/S3Errors.scala
```diff
@@ -46,6 +46,11 @@ object S3Errors {
     case exc: S3Exception if exc.statusCode() == 500 =>
       new StoreWriteError(exc) with RetryableError
 
+    // e.g. S3Exception: Object key is too long. Maximum number of bytes allowed in keys is 915.
+    case exc: S3Exception
+        if exc.getMessage.startsWith("Object key is too long") =>
+      InvalidIdentifierFailure(exc)
```
alexwlchan, line 52: For some reason, we used to be doing this ourselves – see the code that handles `MAX_KEY_BYTE_LENGTH`. The SDK will check this for us, so I see no reason to duplicate that logic.

storage/src/main/scala/weco/storage/store/s3/S3MultipartUploader.scala
```diff
@@ -0,0 +1,76 @@
...
+import grizzled.slf4j.Logging
+import software.amazon.awssdk.core.sync.RequestBody
+import software.amazon.awssdk.services.s3.S3Client
+import software.amazon.awssdk.services.s3.model.{CompleteMultipartUploadRequest, CompletedMultipartUpload, CompletedPart, CreateMultipartUploadRequest, UploadPartRequest}
+import weco.storage.s3.S3ObjectLocation
+
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+trait S3MultipartUploader extends Logging {
```
alexwlchan, line 18: I've pulled out these helper functions so we can use them in two places:

1. In S3StreamWritable, where we know how long every part is going to be in advance, and we say upfront "I need N parts of size S"
2. In the snapshot generator, where we're just getting a stream of gzip-compressed JSON bytes, and we don't know in advance how big it'll be

Also we may want to wrap some of these functions in retrying logic later. I've not done that on the first pass, because we don't have any in the snapshot generator and it's been fine, and other layers of retrying may catch it (e.g. at the SQS level), but noting as a future option.

storage/src/main/scala/weco/storage/store/s3/S3StreamWritable.scala
```diff
@@ -1,101 +1,92 @@
...
+      uploadPart(location, uploadId, bytes, partNumber)
+    }.toList
+
+    val successes = result.collect { case Success(s) => s }
+    val failures = result.collectFirst { case Failure(e) => e }
+
+    failures match {
+      case None    => Success(successes)
+      case Some(e) => Failure(e)
+    }
```
alexwlchan, line 81: Theoretically we could bail out at the first failure, but I expect failures to be rare here.

jamieparkinson (approved):

PR #214, opened 2023-05-26
try removing the aws.crt library

alexwlchan:
This is causing storage service issues and I don't think we need it, so let's try removing it and see if that unbreaks things.


jamieparkinson (approved):
Is this because we all use LCD monitors?

PR #245, opened 2024-07-05
Update Alpakka to version 4.0.0 (#2678)

StepanBrychta:
## What does this change?

Bump Alpakka version to 4.0.0. This is the first step towards migrating from Akka to Pekko — Pekko Connectors is a replacement for Alpakka but it's based on version 4.0.0, so we need to upgrade first. See https://pekko.apache.org/docs/pekko-connectors/1.0/release-notes/index.html.

Updating Alpakka to version 4.0.0 requires:
* Updating Scala from 2.12 to 2.13 (Alpakka 4.0.0 [only supports Scala 2.13](https://mvnrepository.com/artifact/com.lightbend.akka/akka-stream-alpakka-google-cloud-pub-sub-grpc))
* Updating the Scanamo package to successfully resolve dependencies

Alpakka is successfully updated to version 4.0.0 and no bugs are introduced in the process.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,8 @@
+RELEASE_TYPE: minor
+
+    Update Scala to version 2.13.14
+    Update Scanamo to version 1.0.4
+    Update Alpakka to version 4.0.0
+    See https://github.com/wellcomecollection/catalogue-pipeline/issues/2678 for more information
+    
+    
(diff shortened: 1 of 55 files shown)


StepanBrychta:
See https://wellcome.slack.com/archives/C02ANCYL90E/p1720454081908029

PR #250, opened 2024-09-17
Publish packages to Sonatype

StepanBrychta:
## What does this change?

* Create a new GitHub workflow which publishes scala-libs packages to Sonatype.
* Create two other GitHub workflows for testing and reporting evictions, replacing the original Buildkite workflow


project/Common.scala
```diff
@@ -1,13 +1,29 @@
...
+      )
+    ),
+    developers ++= List(
+      Developer(
+        id    = "sbrychta",
+        name  = "Stepan Brychta",
+        email = "s.brychta@wellcome.org",
+        url   = url("https://github.com/StepanBrychta")
+      )
+    ),
```
StepanBrychta, line 26: Sonatype [requires a developers section for all published packages](https://central.sonatype.org/publish/requirements/#developer-information).

Do we want to add all current contributors here? And what about past contributors?

.github/workflows/release.yml
```diff
@@ -0,0 +1,110 @@
...
+          - sierra
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Checkout the latest version, which includes the changes pushed by the previous step!
+          # If we didn't do this, we would be publishing using the previous version tag.
+          ref: main
+      - name: Set up GPG
+        run: |
+          echo "${{ secrets.BASE64_GPG_KEY }}" | base64 -d > secret-keys.gpg
```
kenoir, line 94: You might elsewhere (in a different stack) be able to terraform these values on to the GitHub repository so there existence is codified and updated values in GitHub can be done with a terraform apply.

.github/workflows/report-evictions.yml
```diff
@@ -0,0 +1,74 @@
...
+      # Post the eviction report as a comment on the corresponding PR
+      - name: Collate evictions
+        run: |
+          mkdir -p .reports
+          builds/report_unique_evictions.sh | tee unique_evictions.txt
+      - name: Post eviction report comment
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const postComment = require('./.github/scripts/post-eviction-report-comment.js');
```
kenoir, line 73: This is great.

project/Common.scala
kenoir: I don't mind this being incomplete/inaccurate. I don't think this has to be a real person and could probably refer to the "Wellcome Collection" as a whole?

kenoir (approved):
This is really fantastic work. 

Only comment is it might be good to add some of the context about what's going on to a README somewhere, and mention the Sonatype account setup too.

.github/workflows/release.yml
```diff
@@ -0,0 +1,110 @@
...
+  # Publishing involves several steps:
+  # - Configuring a GPG key so that the packages can be signed
+  # - Configuring Sonatype credentials
+  # - Publishing the packages to a local staging repository using the sbt-sonatype plugin
+  # - Releasing the published bundle to Sonatype
+  publish:
+    runs-on: ubuntu-latest
+    needs: [create-release, check-for-release-file]
+    if: needs.check-for-release-file.outputs.has-release == 'true'
+    strategy:
```
agnesgaroux, line 70: Since `create-release` job needs `check-for-release-file` and won't run unless `has-release == 'true'` , would it work here with only
`needs: create-release`?

.github/workflows/release.yml
agnesgaroux: [Here](https://github.com/wellcomecollection/aws-account-infrastructure/blob/main/accounts/platform/gha_org_secrets.tf) a GHA secret is added at the org level and its access configured. Its value is also created by the tf so we can have it as `module.gha_scala_formatting_role.role_arn`

.github/workflows/release.yml
StepanBrychta: You're right, I simplified the logic

project/Common.scala
StepanBrychta: Done, I replaced my details with general Wellcome Collection details

kenoir (approved):

kenoir:
Part of https://github.com/wellcomecollection/platform/issues/5783

PR #251, opened 2024-09-23
Give release action permissions to push to main

StepanBrychta:
## What does this change?

Gives the `release` GitHub action the necessary permissions to push changes directly into main. 

This is done by configuring the `release` action to assume the role of the Wellcome Collection GitHub app when pushing changes to main. The implementation involves creating a token from the app's private key and follows [official GitHub documentation](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow).

Separately, the Wellcome Collection GitHub app was given the necessary permissions to bypass the [requires-review](https://github.com/organizations/wellcomecollection/settings/rules/638886) rule.

## How to test

I tested this change by temporarily removing the part of the workflow which publishes to Sonatype and running the GitHub action from this branch, which resulted in a successful push to main (see [here](https://github.com/wellcomecollection/scala-libs/commit/3ae383e8ea7050b126c345a6edf93130b4097807)), which I subsequently reverted. 

The `release` action runs successfully. 

Giving the GitHub action full repo access comes with security risks. However, the original Buildkite pipeline was using a similar setup, so this should not introduce any new risks.

I have considered alternative release setups which do not involve pushing directly into main, but these setups have significant limitations (see [here](https://wellcome.slack.com/archives/C02ANCYL90E/p1726759356031529)) and are not viable alternatives.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,13 +45,21 @@ jobs:
           git fetch --tags
           LATEST_TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
           python3 .github/scripts/create_release.py ${LATEST_TAG} $(pwd)
+
+      - name: Generate a token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: 129326 # App ID of the Wellcome Collection app
+          private-key: ${{ secrets.WELLCOME_COLLECTION_APP_PRIVATE_KEY }}
+
       - name: Configure git
         # We need to give the GitHub action full repo privileges via a PAT so that it can push the release directly into main
         run: |
           git config --global user.name "GitHub on behalf of Wellcome Collection"
           git config --global user.email "wellcomedigitalplatform@wellcome.ac.uk"
           
-          git remote set-url origin https://x-access-token:${{ secrets.FULL_REPO_PRIVILEGES_PAT }}@github.com/${{ github.repository }}.git
+          git remote set-url origin https://x-access-token:${{ steps.generate-token.outputs.token }}@github.com/${{ github.repository }}.git
 
       - name: Commit and push changes
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,24 +38,32 @@ jobs:
     if: needs.check-for-release-file.outputs.has-release == 'true'
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - name: Update CHANGELOG.md and build.sbt
         run: |
           git fetch --tags
           LATEST_TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
           python3 .github/scripts/create_release.py ${LATEST_TAG} $(pwd)
-      - name: Commit and push changes
-        env:
-          GITHUB_TOKEN: ${{ secrets.FULL_REPO_PRIVILEGES_PAT }}
+      - name: Configure git
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -38,22 +38,33 @@ jobs:
     if: needs.check-for-release-file.outputs.has-release == 'true'
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - name: Update CHANGELOG.md and build.sbt
```
agnesgaroux, line 43: What does this do?

.github/workflows/release.yml
```diff
@@ -38,22 +38,33 @@ jobs:
...
-      - name: Commit and push changes
+      - name: Configure git
+        # We need to give the GitHub action full repo privileges via a PAT so that it can push the release directly into main
         run: |
-          NEW_TAG=$(cat CHANGELOG.md | grep -m1 -o 'v[0-9]\+\.[0-9]\+\.[0-9]\+')
           git config --global user.name "GitHub on behalf of Wellcome Collection"
           git config --global user.email "wellcomedigitalplatform@wellcome.ac.uk"
+          
+          git remote set-url origin https://x-access-token:${{ secrets.FULL_REPO_PRIVILEGES_PAT }}@github.com/${{ github.repository }}.git
+
```
agnesgaroux, line 55: Who's the "owner" of the PAT? I thought PAT were created on a user so I'm wondering who/what is authenticating here

.github/workflows/release.yml
StepanBrychta: By default, the checkout step persists GitHub credentials, which makes any subsequent steps in the action assume these credentials (even if alternative credentials are specified later on). But since these credentials don't have the necessary permissions, we want them not to persist.

.github/workflows/release.yml
StepanBrychta: Yes that's correct, the PAT was generated by me and since I have admin permissions on the repo, these permissions get transferred onto the PAT, allowing the GitHub action to bypass the "requires-review" rule.

As an alternative (and probably better!) solution, I'm now experimenting with creating a GitHub app for the action, giving it the required permissions, and adding it to the bypass list for the "requires-review" rule.

.github/workflows/release.yml
agnesgaroux: Thanks, makes sense!

.github/workflows/release.yml
agnesgaroux: There's a Wellcome Collection Github app that I don't think is actually used (there's one private key that's never been used). Maybe we can start using it for any automated task that need some permissions

agnesgaroux: https://github.com/organizations/wellcomecollection/settings/apps/wellcome-collection

StepanBrychta: I think this is a good idea. I pushed some new changes, which remove the PAT and instead use the credentials of the Wellcome Collection app.

agnesgaroux (approved):

agnesgaroux:
Very nice! 🙌

agnesgaroux:
Would it be safe to store this private key as an organisation GHA secret? that way we can let more repos use it as needed, but maybe it's safer to have different keys for different repos or even tasks

StepanBrychta:
@agnesgaroux Good idea, the key is now an organisation secret

PR #252, opened 2024-10-01
Improve Sonatype release action #5780

StepanBrychta:
## What does this change?

Make several improvements to the `release` action:
* Increase connection timeout to prevent the action from failing due to timeouts while checking the status of the published package. These timeouts are the result of a known bug in the publishing library we're using (see [here](https://github.com/xerial/sbt-sonatype/issues/518)). Increasing the timeout is a workaround suggested in the comments, but I'll keep an eye on the open issue and update the package if the issue gets fixed.
* Use the Sonatype API to check if the package we are about to upload already exists. If so, exit gracefully.

## How to test

The only major change is in the 'Publish to Sonatype' shell script, which can be tested locally (note that the line with the `sonatypeBundleRelease` command should be commented out when testing).

No more `release` action failures due to timeout issues.

I believe this change doesn't introduce any new risks.


kenoir (approved):
👌

PR #253, opened 2024-10-03
Update dependencies

StepanBrychta:
## What does this change?

* Update dependencies to resolve a few security alerts. This includes the changes from  https://github.com/wellcomecollection/scala-libs/pull/242 and https://github.com/wellcomecollection/scala-libs/pull/241 so that a new version of scala-libs can be released with the updated dependencies. 
* Add the `sierra_typesafe` project into all actions (test, report evictions, and release). In the original version this project was left out by accident (project names were copied from the "Test" group of the original Buildkite pipeline, which also omitted this project).

This change should fix all logback-related security alerts across all repos using `scala-libs`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -107,6 +107,7 @@ jobs:
           - elasticsearch
           - elasticsearch_typesafe
           - sierra
+          - sierra_typesafe
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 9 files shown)


paul-butcher (approved):
I've not tried this out, because I know the the proof of this only really comes when this lib gets updated in the applications that use it, and you are about to sort those out.

PR #258, opened 2024-12-19
Update ElasticBuilder configuration

kenoir:
## What does this change?

This change updates ElasticBuilder to generate and accept config case classes to enable pushing typesafe config to the edge of consuming services.

Part of https://github.com/wellcomecollection/catalogue-pipeline/pull/2793

See [tests](https://github.com/wellcomecollection/scala-libs/pull/258/files#diff-56ee74db28e98cebd14c404087b1034c1f343d746086cbc1fabdcb7110050423R27) for examples of use.

> [!Note]
> This change includes a run of `test:scalafmt` from sbt which formatted other files outside where the functional change takes place, the formatting change is in a separate commit.

## How to test

- [ ] Run the tests, do they pass?

Consumers can push typesafe configuration outside of their applications if they want, making it clearer what configuration is drawn from where, and what config things require.

This change continues to support handing in typesafe config directly, and does not change the interface for existing consumers.


elasticsearch_typesafe/src/test/scala/weco/elasticsearch/typesafe/ElasticBuilderTest.scala
```diff
@@ -3,33 +3,81 @@ package weco.elasticsearch.typesafe
...
+      |es.$namespace.host = $host
+      |es.$namespace.apikey = $apiKey
+      |""".stripMargin
+  )
 
-    noException shouldBe thrownBy(
-      ElasticBuilder.buildElasticClient(config, "test")
-    )
+  describe("when an API key is specified") {
+    it("can build and accept ElasticConfigUsernamePassword") {
```
paul-butcher, line 60:
```suggestion
    it("can build and accept ElasticConfigApiKey") {
```

elasticsearch_typesafe/src/test/scala/weco/elasticsearch/typesafe/ElasticBuilderTest.scala
```diff
@@ -3,33 +3,81 @@ package weco.elasticsearch.typesafe
...
+    f"""
+      |es.$namespace.host = $host
+      |es.$namespace.apikey = $apiKey
+      |""".stripMargin
+  )
 
-    noException shouldBe thrownBy(
-      ElasticBuilder.buildElasticClient(config, "test")
-    )
+  describe("when an API key is specified") {
```
paul-butcher, line 59: There should be a third test to cover the dominance of API Key over user/pass when both are provided.

build.sbt
```diff
@@ -1,4 +1,4 @@
-val projectVersion = "32.43.2"
+val projectVersion = "32.43.2-SNAPSHOT"
```
paul-butcher, line 1: Should this get a new version number when it goes in?

paul-butcher (approved):

build.sbt
kenoir: I believe the build [process does that](https://github.com/wellcomecollection/scala-libs/blob/22d55dd34350ed4604e563bebfc3a146ab8a5050/.github/scripts/create_release.py#L63) as part of the release.

PR #260, opened 2025-01-06
Use localstack in storage tests

kenoir:
## What does this change?

Follows https://github.com/wellcomecollection/scala-libs/pull/259

Instead of using the Zenko version of S3, this switches to using the Localstack version. 

We may wish to use Localstack to avoid future maintenance issues with the Zenko images that don't seem to have been updated in the last 2 years, and so that we can use the ECR public versions of the Localstack images to avoid potential throttling on docker hub (which has been an issue in the past.

> [!Note]
> This change required updated the tests as some of the error messages emitted by the Localstack S3 container image around bucket naming are different from the Zenko image. I have not determined which is "correct" as per comparison to S3 but allowed both.

## How to test

- [ ] Run the tests, do they pass?

Tests can pass, we can release things.

The differing errors may mean that Localstack is incorrect when compared to S3, we've kept both old and new error messages as valid to mitigate.

### DIFF
--- a/storage/docker-compose.yml
+++ b/storage/docker-compose.yml
@@ -5,19 +5,9 @@ services:
     ports:
       - "45678:8000"
   s3:
-    image: "zenko/cloudserver:8.1.8"
+    image: "public.ecr.aws/localstack/localstack:4.0.0"
     environment:
-      - "S3BACKEND=mem"
+      - SERVICES=s3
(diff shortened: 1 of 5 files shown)


paul-butcher (approved):
Yes.  This is a much better idea than my fix

PR #263, opened 2025-01-07
Setup sbt in release action

kenoir:
## What does this change?

This change adds a GitHub action step to ensure sbt is setup as part of the release action, as this is currently failing: https://github.com/wellcomecollection/scala-libs/actions/runs/12636383833/job/35209020492#step:6:27

I believe there has been a change to what is available on runners (sbt was previously available by default on GHA runners, it now is not and needs explicitly adding). See: https://github.com/actions/setup-java/issues/266

In the future we could consider having our own action for this in the style of https://github.com/guardian/setup-scala (credit to the great @rtyley).

## How to test

- [ ] Merge and release, is it successful?

We can release our shared libs!

This shouldn't break anything further!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -126,11 +126,13 @@ jobs:
           mkdir ~/.sbt
           echo "${{ secrets.SONATYPE_CREDENTIALS }}" > ~/.sbt/sonatype.credentials
       - name: Setup JDK
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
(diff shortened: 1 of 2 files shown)


agnesgaroux (approved):

rtyley:
> In the future we could consider having our own action for this in the style of https://github.com/guardian/setup-scala 

Absolutely! Feel free to fork `guardian/setup-scala` or even use it directly if you don't mind using `.tool-versions` etc.

